PR #14, opened 2021-05-05
Do not leak the global snapshot listener

JakeWharton:


JakeWharton:
6 months of self-inflicted bugs to fix. PR 1 of 352...

PR #37, opened 2021-06-26
Update Usage

dcampogiani:
replace `org.jakewharton.mosaic` with `com.jakewharton.mosaic`


JakeWharton:
Thanks. Shouldn't write documentation late at night.

PR #54, opened 2022-08-10
Update to kotlin 1.7.10

saket:
I ran all the samples and they look fine.

### DIFF
--- a/mosaic/build.gradle
+++ b/mosaic/build.gradle
@@ -1,3 +1,6 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile
+
 buildscript {
 	apply from: '../dependencies.gradle'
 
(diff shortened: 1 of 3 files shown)


mosaic/build.gradle
```diff
@@ -60,10 +63,9 @@ allprojects {
 		}
 	}
 
-	tasks.withType(org.jetbrains.kotlin.gradle.dsl.KotlinJvmCompile).configureEach { task ->
+	tasks.withType(KotlinJvmCompile).configureEach { task ->
```
saket, line 66: IntelliJ was showing yellow lines for these fully qualified names. I can undo these imports if you want.

JakeWharton:
We can actually nuke the submodule and switch to JetBrains' runtime like we do in Molecule and Redwood! I'll do in a follow-up.

saket:
that'd be awesome!

JakeWharton:
Tracked by #55

PR #66, opened 2022-08-19
Add Static component for rendering permanent output

### LINKED ISSUE #65: Mixing console logging with mosaic
I'd like to create something similar to the rich console output of Gradle with Mosaic, where some kind of table is pinned to the bottom and log messages append above. This is currently doable within Mosaic if I keep a list of these log messages as state and display them first in the content and then the table. But this seems really inefficient as both the log messages and the table are completely cleared from the console and re-written with each update.

What I think I'm looking for is some kind of `Appender` which is available through `MosaicScope` that I can call something like `println` on. The internal `AnsiOutput` will know how to write those lines after clearing the screen of the previous output but before writing the new output, thus keeping the `setContent` at the bottom of the console. But this is only what I think and there's probably some better way of doing this.

I'd be happy to mess around with this myself if you like the idea and would be open to contribution.

bnorm:
Resolves #65 

Probably went a little overboard on the `jest` sample updates so let me know if you want me to roll anything back. Also I'm not stuck on any naming, so if you have suggestions let me know.


mosaic-runtime/src/main/kotlin/com/jakewharton/mosaic/Static.kt
```diff
@@ -0,0 +1,108 @@
...
+ * actually been written to a [TextCanvas], the [postRender] callback will be
+ * invoked to allow clearing of content.
+ *
+ * @param postRender Callback after rendering to a [TextCanvas] is complete.
+ * @param content Content which should be rendered permanently above normal
+ * canvas.
+ */
+@Composable
+internal fun Static(
+	postRender: @DisallowComposableCalls () -> Unit = {},
```
JakeWharton:
```suggestion
	postRender: () -> Unit = {},
```
This should only be needed if the enclosing function was inline.

mosaic-runtime/src/main/kotlin/com/jakewharton/mosaic/Static.kt
```diff
@@ -0,0 +1,108 @@
...
+		// Render contents of static node to a separate display.
+		val other = box.render()
+
+		// Add display canvas to static canvases if it is not empty.
+		if (other.width > 0 && other.height > 0) {
+			canvas.static.add(other)
+		}
+
+		// Propagate any static content of the display.
+		canvas.static.addAll(other.static)
```
JakeWharton, line 102: This is my least favorite part of the change. The fact that a render causes a side-effect to collect these other renders and then something at the call site magically emits that first.

It's hard to argue with the output, but I think I would like to try and find something better for this. I'm not going to block merging here because the output is truly fantastic and it's all implementation detail.

Render (well, renderTo) is a tree walk, and I think perhaps we separate out the static tree walk to its own function which can be less reliant on side-effects + mutation. It also hopefully means the canvas isn't aware of these static renders.

JakeWharton (approved):
Super slick!

mosaic-runtime/src/main/kotlin/com/jakewharton/mosaic/Static.kt
```diff
@@ -0,0 +1,108 @@
...
+	LaunchedEffect(items) {
+		items.collect {
+			pending.add(Item(it, rendered = false))
+		}
+	}
+
+	Static(
+		postRender = {
+			// Remove any items which have been rendered.
+			pending.removeIf { it.rendered }
```
JakeWharton, line 35: This isn't super efficient but it's probably fine. One thing to note is that the `true`s are contiguous starting at index 0 so we could find the first `false` index and then do `subList(0, firstFalse).clear()` for an operation that is likely the most optimal.

JakeWharton:
Slamming it home. We can iterate as needed. Thanks for this!

mosaic-runtime/src/main/kotlin/com/jakewharton/mosaic/Static.kt
bnorm: Yeah, this was definitely a result of not wanting to make a bunch of render things `Static` specific or introduce multiple render passes. Thankfully as you said it is all implementation detail so can definitely be iterated.

> Render (well, renderTo) is a tree walk, and I think perhaps we separate out the static tree walk to its own function which can be less reliant on side-effects + mutation.

Are you thinking something like 2 `renderTo` passes, one for static and the other for "normal" content? Or find all the `Static` nodes in the `render` function and render those to the canvas first somehow? I can try a couple different things - if you want - but maybe a little pseudocode of what you are thinking might help.

> It also hopefully means the canvas isn't aware of these static renders.

The tricky part here was that `AnsiOutput` needs to know the static and non-static parts so it can set the `lastHeight` to only the non-static part. So whatever gets sent to the output, needs to know that additional detail.

mosaic-runtime/src/main/kotlin/com/jakewharton/mosaic/Static.kt
bnorm: That's a good idea. I always forget that modifications on subList are reflected in the full list. Also worth noting that just `pending.clear()` worked in all of the test samples I created but I wasn't convinced that would work in all cases so I added the wrapper `Item` class to be safe.

mosaic-runtime/src/main/kotlin/com/jakewharton/mosaic/Static.kt
JakeWharton: I was thinking a whole separate function, like `renderTo()` and `renderStaticTo()` or something. Basically just splitting the responsibilities and each node would only realistically implement one of them.

PR #68, opened 2022-08-28
Migrate to version catalogs

### LINKED ISSUE #59: Migrate to version catalogs

danherrera:
Closes #59


build.gradle
```diff
@@ -1,10 +1,10 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 buildscript {
-	apply from: 'dependencies.gradle'
 
 	dependencies {
-		classpath deps.kotlinGradlePlugin
+		def libs = project.extensions.getByType(VersionCatalogsExtension).named("libs")
+		classpath libs.findLibrary("kotlin-gradlePlugin").get()
```
JakeWharton: Why this indirection? And why not include AGP? We have projects with version catalogs that use the libraries section to set up the classpath of the buildscript rather than using the plugin ID.

danherrera: Oh, interesting. I thought there was a limitation that prevented the buildscript dependencies from directly accessing **libs**. Good to know that isn't the case!

I based my changes on the dependencies in `dependencies.gradle` and missed these! They should be updated now. Thanks!

danherrera: > We have projects with version catalogs that use the libraries section to set up the classpath of the buildscript rather than using the plugin ID

Is this referring to the use of `libs.kotlin.gradlePlugin` instead of `libs.findLibrary("kotlin-gradlePlugin")` or something else?

build.gradle
```diff
@@ -1,16 +1,15 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 buildscript {
-	apply from: 'dependencies.gradle'
```
JakeWharton:
```suggestion
```

mosaic-gradle-plugin/src/test/fixture/counter/build.gradle
```diff
@@ -1,8 +1,7 @@
 buildscript {
-	apply from: "../../../../../dependencies.gradle"
```
JakeWharton:
```suggestion
```

samples/build.gradle
```diff
@@ -1,8 +1,7 @@
 buildscript {
-	apply from: '../dependencies.gradle'
```
JakeWharton:
```suggestion
```

JakeWharton (approved):
Thanks!

PR #70, opened 2022-09-02
Fewer lines of output will cause display to drift downwards

bnorm:
Instead of opening an issue, I thought this might be easier to demonstrate, so went straight for the PR.

[I believe the `lastHeight` logic of `AnsiOutput` to be slightly flawed](https://github.com/JakeWharton/mosaic/blob/trunk/mosaic-runtime/src/main/kotlin/com/jakewharton/mosaic/output.kt#L33). I've tested this with a modified version of the jest sample which does not display the done tests. I found that if the new output is smaller by 2 or more lines, the output will drift downwards and not clear all previous output. I've included an SVG in the first commit along with the test change which demonstrates this bug.

I found that the extra line clearing logic leaves the cursor too far down which causes the initial cursor move up on the next frame to not move far enough and leave some of the previous output. If the cursor is moved back up after clearing extra lines, all of the previous output is cleared correctly. I've made this change in the second commit and included another SVG of the jest sample output.

The third commit then reverts all the jest sample changes so this PR could be squash-merged and only include the minimal number of changes.

Another possible solution is instead of moving the cursor up, add `minOf(0, extraLines - 1)` to `lastHight` in addition to `lines.size`. This however leaves cleared lines below the output when the program exits where moving the cursor up will not. But due to the extra line clearing having a conditional new line, the cursor will drift back up slowly each frame which makes for a slightly odd visual. I think overall moving the cursor up after the extra line clearing is better visually and logically.

Moving cursor up:
```text
$ ./build/install/jest/bin/jest
Tests: 4 failed, 6 passed, 10 total
Time:  10s

$ # next prompt here
```

Adding extra lines to last hight:
```text
$ ./build/install/jest/bin/jest
Tests: 4 failed, 6 passed, 10 total
Time:  10s


$ # next prompt here
```


mosaic-runtime/src/main/kotlin/com/jakewharton/mosaic/output.kt
```diff
@@ -47,13 +47,19 @@ internal object AnsiOutput : Output {
...
-			for (i in 0 until lastHeight - lines.size) {
+			val extraLines = lastHeight - lines.size
+			for (i in 0 until extraLines) {
 				if (i > 0) {
 					append('\n')
 				}
 				append("\u001B[K") // Clear line.
 			}
 
+			// Move cursor back up to end of the new output
```
JakeWharton:
```suggestion
			// Move cursor back up to end of the new output.
```
nit. we follow https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

JakeWharton (approved):

JakeWharton:
Thanks!

PR #156, opened 2023-03-14
Render nodes in debug output even if drawing fails

### LINKED ISSUE #157: Measure and place static branches with main tree
Right now we defer static measure and placement until we call `drawStatics()` which does all three operations. `draw()` performs all three operations on the main tree so it seems symmetric, but it's actually not.

JakeWharton:
Closes #157


JakeWharton:
Good test

PR #211, opened 2023-08-11
Fix gradle extension name in README

ZacSweers:


JakeWharton (approved):
I thought I fixed all these. Total copy/pasta job. I had the KDoc wrong too!

PR #226, opened 2023-10-07
Update robot example

EpicDima:
Use padding modifier instead of building custom string for Text

### DIFF
--- a/samples/robot/src/main/kotlin/example/robot.kt
+++ b/samples/robot/src/main/kotlin/example/robot.kt
@@ -3,6 +3,8 @@ package example
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
+import com.jakewharton.mosaic.layout.padding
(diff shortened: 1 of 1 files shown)


samples/robot/src/main/kotlin/example/robot.kt
```diff
@@ -24,14 +26,7 @@ fun main() = runMosaicBlocking {
...
 			// TODO https://github.com/JakeWharton/mosaic/issues/11
-			Text(buildString {
-				// TODO https://github.com/JakeWharton/mosaic/issues/7
-				repeat(y) { append('\n') }
-				repeat(x) { append(' ') }
-				append("^_^")
-
-				repeat(height - y) { append('\n') }
-			})
+			Text("^_^", modifier = Modifier.padding(x, y, width - x, height - y))
```
EpicDima, line 29: Padding on the right is not really necessary, but without it world will not be beautiful if it's colored, like this
image

JakeWharton:
Excellent! Thanks

PR #227, opened 2023-10-11
Add AnnotatedString with SpanStyle for text customization

### LINKED ISSUE #9: AnnotatedString
I should be able to control styles in a single `Text` without needing many of them

EpicDima:
This is a simplified version of AnnotatedString and SpanStyle from Jetpack Compose. There is no ParagraphStyle support. SpanStyle contains fewer properties. The API has been saved for using AnnotatedString and creating and styling it, for quick understanding if you know Jetpack Compose.

The tests are also taken from Jetpack Compose, but simplified.

The code has two Compose Text functions, two DrawScope.drawText methods, two TextLayout inheritors, in which the difference is only in the accepted value - either String or AnnotatedString. This is done for optimization, but in theory it may be unnecessary.

Update the jest example using AnnotatedString.

This is an attempt to implement a full-featured AnnotatedString, but with the functionality currently available 🙂

Closes #9

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawScope.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawScope.kt
@@ -17,6 +17,9 @@
 package com.jakewharton.mosaic.layout
 
 import com.jakewharton.mosaic.TextCanvas
+import com.jakewharton.mosaic.text.AnnotatedString
+import com.jakewharton.mosaic.text.SpanStyle
+import com.jakewharton.mosaic.text.getLocalRawSpanStyles
 import com.jakewharton.mosaic.ui.Color
 import com.jakewharton.mosaic.ui.TextStyle
 
@@ -40,13 +43,22 @@ public interface DrawScope {
 		background: Color? = null,
 		style: TextStyle? = null,
 	)
+
+	public fun drawText(
+		row: Int,
+		column: Int,
+		string: AnnotatedString,
+		foreground: Color? = null,
+		background: Color? = null,
+		style: TextStyle? = null,
+	)
 }
 
 internal open class TextCanvasDrawScope(
 	private val canvas: TextCanvas,
 	override val width: Int,
 	override val height: Int,
-): DrawScope {
+) : DrawScope {
 	override fun drawRect(
 		color: Color,
 		row: Int,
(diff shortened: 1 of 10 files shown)


EpicDima:
To show how it works, I modified the counter example with the following code:
```kotlin
setContent {
  Text(
    buildAnnotatedString {
      withStyle(SpanStyle(Color.Blue)) {
        append("The")
      }
      append(" ")
      withStyle(SpanStyle(Color.Red)) {
        append("count")
      }
      append(" ")
      withStyle(SpanStyle(Color.BrightGreen)) {
        append("is:")
      }
      addStyle(SpanStyle(Color.BrightMagenta), 3, 11)
      addStyle(SpanStyle(background = Color.BrightRed), 7, 15)
      addStyle(SpanStyle(textStyle = TextStyle.Underline), 1, 8)
      append(" $count")
    }
  )
}
```

Result:
image

JakeWharton:
Thanks! I will look soon

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawScope.kt
```diff
@@ -40,13 +43,22 @@ public interface DrawScope {
...
 		style: TextStyle? = null,
 	)
+
+	public fun drawText(
+		row: Int,
+		column: Int,
+		string: AnnotatedString,
+		foreground: Color? = null,
+		background: Color? = null,
+		style: TextStyle? = null,
```
JakeWharton, line 53: Do we need these? Should we get rid of them on both overloads?

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/text/TextLayout.kt
```diff
@@ -2,30 +2,31 @@ package com.jakewharton.mosaic.text
 
 import de.cketti.codepoints.codePointCount
 
-internal class TextLayout {
-	var value: String = ""
+public abstract class TextLayout<T : CharSequence>(initialValue: T) {
```
JakeWharton: I can't see why this and the subtypes need to be public.

JakeWharton (approved):

JakeWharton:
Once this lands, we can migrate the types which define `equals` and `hashCode` to use [Poko](https://github.com/drewhamilton/Poko).

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawScope.kt
EpicDima: If I understood correctly, then the question is about default values that we really don't use and probably they can be removed

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/text/TextLayout.kt
EpicDima: Thanks, fixed it!

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawScope.kt
JakeWharton: I mean the arguments altogether. They seem like a redundant codepath now that you can style within the annotated string. Moreover, they create a precedence problem. If you specify foreground here and as a span which wins?

However, when I look at Compose UI, it seems like they have _three_ colors. You can specify it on a span, on the text as an argument, or on the text within a text style. Ugh. I guess we leave it as-is for now.

PR #229, opened 2023-10-15
Add MosaicComposable annotation

### LINKED ISSUE #51: Adopt target markers in generated Compose code
https://developer.android.com/jetpack/androidx/releases/compose-runtime#1.2.0-alpha04

This will ensure no one uses our Compose code in Compose UI or other compositions.

EpicDima:
The annotation is used to mark Mosaic Compose functions that cannot be used in other Compose code or for other compositions.

Closes https://github.com/JakeWharton/mosaic/issues/51


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/MosaicComposable.kt
```diff
@@ -0,0 +1,21 @@
+package com.jakewharton.mosaic.ui
```
EpicDima, line 1: It may need to be moved to another package

JakeWharton:
Thanks!

PR #75, opened 2022-09-06
WIP Expose TerminalInfo

JakeWharton:
Currently breaks Robot sample input since it also wants to init Jline

#72

Issue #133, opened 2023-02-28
Add `Constraints` to layout system


JakeWharton:
We'll eventually be able to pipe down the terminal size

PR #230, opened 2023-10-16
CompositionLocal for terminal size by Mordant

### LINKED ISSUE #72: `CompositionLocal` for terminal size
Feature request extracted from #41

EpicDima:
The Mordant KMP library has been added to the mosaic-runtime module in commonMain, which allows you to interact with the cli.

This makes it possible to get the size of the terminal via `CompositionLocal` - `Terminal.current.size`.

I think it was **not** the best solution for https://github.com/JakeWharton/mosaic/issues/72 :(

Advantages:
- we use Mordant - this is a KMP library, that is, we quickly get a working solution with the size of the terminal
- it is possible to replace `platformDisplay` with `com.github.ajalt.mordant.terminal.Terminal::rawPrint`
- probably now will help to implement Constraints (https://github.com/JakeWharton/mosaic/issues/133 )

Disadvantages:
- another loop is added to check the size of the terminal, instead of using callbacks, which are not in the library
- in the future, it will not be possible to make `onKeyEvent`, because the library does not provide such an API
- in the future, we will probably have to switch to our own solution for updating the size of the terminal via callback and the same for `onKeyEvent` (it's also interesting to think about text input)

I think such a solution may be enough at the moment, since the size of the terminal seems to be very necessary for further development, and working with it in general for Compose code via CompositionLocal will not differ depending on the provider.

Part of the idea is taken from https://github.com/JakeWharton/mosaic/pull/75 😊

Closes #75. Closes #72.

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
@@ -4,11 +4,13 @@ import androidx.compose.runtime.AbstractApplier
 import androidx.compose.runtime.BroadcastFrameClock
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Composition
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.Recomposer
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.snapshots.Snapshot
+import com.github.ajalt.mordant.terminal.Terminal
 import com.jakewharton.mosaic.layout.MosaicNode
 import com.jakewharton.mosaic.ui.BoxMeasurePolicy
-import kotlin.time.ExperimentalTime
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
@@ -17,6 +19,7 @@ import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.yield
+import kotlin.time.ExperimentalTime
 
 /**
  * True for a debug-like output that renders each "frame" on its own with a timestamp delta.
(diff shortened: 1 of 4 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -92,10 +95,35 @@ public suspend fun runMosaic(body: suspend MosaicScope.() -> Unit): Unit = corou
...
+		TerminalInfo(
+			size = TerminalInfo.Size(terminal.info.width, terminal.info.height)
+		)
+	)
+
+	launch(context = composeContext) {
+		while (true) {
+			if (terminal.info.updateTerminalSize()
+				&& (terminalInfo.value.size.width != terminal.info.width
+					|| terminalInfo.value.size.height != terminal.info.height)
```
EpicDima: For `updateTerminalSize` documentation says that it returns `true` if the size has been updated, but in reality it returns `true` even if the size remains the same, so I added checks to avoid creating objects once again

JakeWharton: Should we file a bug on Mordant about that behavior then?

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -92,10 +95,35 @@ public suspend fun runMosaic(body: suspend MosaicScope.() -> Unit): Unit = corou
...
+			) {
+				terminalInfo.value = terminalInfo.value.copy(
```
JakeWharton:
```suggestion
      val currentTerminalInfo = terminalInfo.value
			if (terminal.info.updateTerminalSize()
				&& (currentTerminalInfo.size.width != terminal.info.width
					|| currentTerminalInfo.size.height != terminal.info.height)
			) {
				terminalInfo.value = currentTerminalInfo.copy(
```
Multiple calls to `.value` cause more work than one might imagine to occur. Doing it once will save that work, especially for something happening 20 times per second.

JakeWharton (approved):

JakeWharton:
Similar to your other PR, we'll eliminate the data classes for Poko once this lands.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
EpicDima: In the implementation, this is how it is intended, but the documentation is a little unclear, so I created an issue https://github.com/ajalt/mordant/issues/135 with a proposal to correct it

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
EpicDima: Oh, thanks, I didn't think(

EpicDima:
> Similar to your other PR, we'll eliminate the data classes for Poko once this lands.

I wonder why a replacement for Poko is needed?)

PR #248, opened 2023-11-13
Fix Gradle warnings

TWiStErRob:
See commit message details for what warning each commit fixes.

This should result in a clean (no-output) `gradlew build` now, except maybe:
```
> Task :kotlinNpmInstall
warning Ignored scripts due to flag.
```
Tracking issue: https://youtrack.jetbrains.com/issue/KT-52578, I'm not sure what side-effect the workaround may have.

### DIFF
--- a/samples/demo/build.gradle
+++ b/samples/demo/build.gradle
@@ -2,4 +2,6 @@ apply plugin: 'org.jetbrains.kotlin.jvm'
 apply plugin: 'com.jakewharton.mosaic'
 apply plugin: 'application'
 
-mainClassName = 'example.DemoKt'
+application {
+	mainClass = 'example.DemoKt'
+}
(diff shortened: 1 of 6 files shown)


samples/demo/build.gradle
```diff
@@ -2,4 +2,6 @@ apply plugin: 'org.jetbrains.kotlin.jvm'
 apply plugin: 'com.jakewharton.mosaic'
 apply plugin: 'application'
 
-mainClassName = 'example.DemoKt'
+application {
+	mainClass = 'example.DemoKt'
+}
```
TWiStErRob, line 7: if you want to keep single-line, this should be an alternative syntax:
```suggestion
application.mainClass = 'example.DemoKt'
```

PR #250, opened 2023-11-13
Allow the robot to move on Windows

### LINKED ISSUE #97: Robot sample problem on Windows
I have built the latest version from repository and running the robot sample in a command prompt on Windows 11 results in the following output:

```
D:\Project\mosaic\samples\robot\build\distributions\robot-0.4.0-SNAPSHOT\robot-0.4.0-SNAPSHOT\bin>robot.bat
Use arrow keys to move the face. Press ΓÇ£qΓÇ¥ to exit.
Position: 0, 0   World: 20, 10

^_^
```

Nothing happens when I press the arrow keys, `q` works for quit.

TWiStErRob:
PR for this: https://github.com/JakeWharton/mosaic/pull/75/files#r1016401541

Fixes #97

No clue why, but this way it moves. 😅

---

Btw, would be nice to have a terminal-sized world for the robot like in https://github.com/JakeWharton/mosaic/pull/75, but I couldn't get it to work with `LocalTerminal`.

### DIFF
--- a/samples/robot/src/main/kotlin/example/robot.kt
+++ b/samples/robot/src/main/kotlin/example/robot.kt
@@ -41,7 +41,7 @@ fun main() = runMosaicBlocking {
 				'q'.code -> break
 				27 -> {
 					when (reader.read()) {
-						91 -> {
+						91, 79 -> {
 							when (reader.read()) {
 								65 -> y = (y - 1).coerceAtLeast(0)
 								66 -> y = (y + 1).coerceAtMost(height)


TWiStErRob:
I think I found why based on [this](https://github.com/jline/jline3/blob/277dd8c34fd933b3c4ed8be3ab773342c14e3e43/terminal/src/main/java/org/jline/terminal/impl/AbstractWindowsTerminal.java#L389).

79 0x4F O
91 0x5B [

Normally it's ESC[..., but on Windows the `[` key code is reserved for `VK_LWIN | 0x5B | Left Windows key` as listed [here](https://learn.microsoft.com/en-gb/windows/win32/inputdev/virtual-key-codes?redirectedfrom=MSDN). So they chose an other key.

All this is a big guess, but it makes sense to me.

PR #254, opened 2023-11-18
Add offset modifier

EpicDima:
robot sample using the offset modifier. I haven't updated it here, as I'm not sure that the honest size of the world and the maximum position of the robot will look nice in animation.

Example with offset
<details>
<summary>The first picture</summary>

image
</details>
<details>
<summary>The second picture</summary>

image
</details>

<details>
<summary>Full code</summary>

```kotlin
package example

import androidx.compose.runtime.getValue
import androidx.compose.runtime.mutableStateOf
import androidx.compose.runtime.setValue
import com.jakewharton.mosaic.layout.background
import com.jakewharton.mosaic.layout.height
import com.jakewharton.mosaic.layout.offset
import com.jakewharton.mosaic.layout.size
import com.jakewharton.mosaic.modifier.Modifier
import com.jakewharton.mosaic.runMosaicBlocking
import com.jakewharton.mosaic.ui.Box
import com.jakewharton.mosaic.ui.Color
import com.jakewharton.mosaic.ui.Column
import com.jakewharton.mosaic.ui.Spacer
import com.jakewharton.mosaic.ui.Text
import com.jakewharton.mosaic.ui.unit.IntOffset
import kotlinx.coroutines.Dispatchers.IO
import kotlinx.coroutines.withContext
import org.jline.terminal.TerminalBuilder

private const val width = 20
private const val height = 10

private const val robotWidth = 3
private const val robotHeight = 1

fun main() = runMosaicBlocking {
  // TODO https://github.com/JakeWharton/mosaic/issues/3
  var x by mutableStateOf(0)
  var y by mutableStateOf(0)

  setContent {
    Column {
      Text("Use arrow keys to move the face. Press “q” to exit.")
      Text("Position: $x, $y   Robot: $robotWidth, $robotHeight   World: $width, $height")
      Spacer(modifier = Modifier.height(1))
      // TODO https://github.com/JakeWharton/mosaic/issues/11
      Box(
        modifier = Modifier
          .size(width, height)
          .background(Color.Green)
          .offset { IntOffset(x, y) }) {
        Text("^_^")
      }
    }
  }

  withContext(IO) {
    val terminal = TerminalBuilder.terminal()
    terminal.enterRawMode()
    val reader = terminal.reader()

    while (true) {
      // TODO https://github.com/JakeWharton/mosaic/issues/10
      when (reader.read()) {
        'q'.code -> break
        27 -> {
          when (reader.read()) {
            91, 79 -> {
              when (reader.read()) {
[...]

Issue #153, opened 2023-03-12
`TextSurface` optimizations


JakeWharton:
* [x] Do not combine lines just to split them again
   * #154
* [x] Do not render lines with a string builder just to append to another string builder
   * #154
* [ ] Do not gather a list of text canvases for static content, append to a single canvas
* [x] Eliminate clipping
   * #152 
* [ ] Width and height should be hints, canvas should probably automatically scale to infinity

JakeWharton:
Closing for now. I'm not super interested in optimizing statics and having an infinitely scalable canvas probably isn't a good idea either.

PR #255, opened 2023-11-18
Change the handling of out-of-bounds canvas

EpicDima:
Example from https://github.com/JakeWharton/mosaic/pull/254

<details>
<summary>The first picture</summary>

Screenshot 2023-11-18 at 18 02 05
</details>

<details>
<summary>The second picture</summary>

Screenshot 2023-11-18 at 18 02 28
</details>

It seems to be the **partial** solution for point 5 in https://github.com/JakeWharton/mosaic/issues/153

Update:
Now this is a simple logic that does not cause crashes, while it allows you to partially draw objects that are partially shifted beyond the border. In the first commit, the canvas expanded automatically, but since this is done in the get method, offset modifier expanded the boundaries, although it should not be as a modifier.

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/surface.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/surface.kt
@@ -18,6 +18,11 @@ import de.cketti.codepoints.appendCodePoint
 
 private val blankPixel = TextPixel(' ')
 
+/**
+ * Returns always a new blank [TextPixel].
+ */
+private inline fun newBlankPixel(): TextPixel = TextPixel(' ')
+
 public interface TextCanvas {
 	public val height: Int
 	public val width: Int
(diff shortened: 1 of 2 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/canvas.kt
```diff
@@ -19,25 +19,58 @@ internal interface TextCanvas {
...
+		val y = translationY + row
+		if (x < 0 || y < 0) {
+			return reusableDirtyPixel
+		}
+		val widthDiff = x - width + 1
+		if (widthDiff > 0) {
+			if (widthDiff == 1) {
+				rows.forEach { it.add(newBlankPixel) }
+			} else {
+				rows.forEach { it.addAll(createBlankRow(widthDiff)) }
```
EpicDima: I think this is not an ideal solution, because several lists are created, but on the other hand it should not be a hot execution path.

JakeWharton:
I would like to write tests for this behavior before committing

JakeWharton:
I haven't forgotten about all these PRs. I'm just busy, but they hover near the top of my "TODO" list and "Feels-bad-about-not-yet-merging" list.

PR #256, opened 2023-11-18
Update robot sample with offset modifier

EpicDima:
Previously, the actual width and height were greater than the specified ones, now exactly the same. That's also why a few clicks are removed from input.sh .

It was:
Screenshot 2023-11-19 at 00 56 35

Become:
Screenshot 2023-11-19 at 00 57 23

### DIFF
--- a/samples/robot/src/main/kotlin/example/robot.kt
+++ b/samples/robot/src/main/kotlin/example/robot.kt
@@ -3,17 +3,25 @@ package example
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
-import com.jakewharton.mosaic.layout.padding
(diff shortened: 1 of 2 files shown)


samples/robot/src/main/kotlin/example/robot.kt
```diff
@@ -23,10 +31,12 @@ fun main() = runMosaicBlocking {
 	setContent {
 		Column {
 			Text("Use arrow keys to move the face. Press “q” to exit.")
-			Text("Position: $x, $y   World: $width, $height")
-			Text("")
+			Text("Position: $x, $y   Robot: $robotWidth, $robotHeight   World: $width, $height")
+			Spacer(Modifier.height(1))
```
JakeWharton, line 35: Nice

JakeWharton (approved):

PR #261, opened 2023-11-27
Unit tests for modifiers and some fixes

EpicDima:
Added unit tests for `AspectRatio`, `Offset`, `Padding`, `Size` modifiers. 
Added unit tests for `Arrangement`.
Added some auxiliary tools for unit tests.
Added `equals` and `hashCode` implementations for `SizeModifier`, `FillModifier`, `WrapContentModifier`, `UnspecifiedConstraintsModifier`. Added a new `Composable` - `Filler`. Draws the same symbol over its entire size. Similar to `Spacer`, where such a symbol is a space without additional styles. 
Added a `Composable` `Box` without content.
Fixed formatting in the `toString` implementation for `Alignment`, taking into account possible non-integers. 
Fixed a problem in `OffsetModifier` with incorrect transmission of `x` and `y` due to shadowing fields.
Fixed some bugs and typos in `Node.kt` related to the `minIntrinsicWidth`, `maxIntrinsicWidth`, `minIntrinsicHeight`, `maxIntrinsicHeight` methods.

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/Size.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/Size.kt
@@ -9,6 +9,7 @@ import com.jakewharton.mosaic.ui.unit.IntSize
 import com.jakewharton.mosaic.ui.unit.constrain
 import com.jakewharton.mosaic.ui.unit.constrainHeight
 import com.jakewharton.mosaic.ui.unit.constrainWidth
+import dev.drewhamilton.poko.Poko
 import kotlin.math.roundToInt
 
 /**
@@ -468,6 +469,7 @@ public fun Modifier.defaultMinSize(
 	minHeight: Int = Unspecified,
 ): Modifier = this.then(UnspecifiedConstraintsModifier(minWidth = minWidth, minHeight = minHeight))
 
+@Poko
 private class SizeModifier(
 	private val minWidth: Int = Unspecified,
 	private val minHeight: Int = Unspecified,
@@ -599,28 +601,6 @@ private class SizeModifier(
 		}
 	}
 
-	override fun equals(other: Any?): Boolean {
-		if (this === other) return true
-		if (other !is SizeModifier) return false
-
-		if (minWidth != other.minWidth) return false
-		if (minHeight != other.minHeight) return false
-		if (maxWidth != other.maxWidth) return false
-		if (maxHeight != other.maxHeight) return false
-		if (enforceIncoming != other.enforceIncoming) return false
-
-		return true
-	}
-
-	override fun hashCode(): Int {
-		var result = minWidth.hashCode()
-		result = 31 * result + minHeight.hashCode()
-		result = 31 * result + maxWidth.hashCode()
-		result = 31 * result + maxHeight.hashCode()
-		result = 31 * result + enforceIncoming.hashCode()
-		return result
-	}
-
 	override fun toString(): String {
 		val params = buildList {
 			if (minWidth != Unspecified) {
@@ -641,6 +621,7 @@ private class SizeModifier(
 	}
 }
 
+@Poko
 private class FillModifier(
 	private val direction: Direction,
 	private val fraction: Float,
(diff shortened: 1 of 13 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/Size.kt
```diff
@@ -599,6 +599,28 @@ private class SizeModifier(
 		}
 	}
 
+	override fun equals(other: Any?): Boolean {
```
JakeWharton: Seems like this could be a `data class` or annotated with `@Poko` to have these two functions generated.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/Size.kt
```diff
@@ -658,6 +680,22 @@ private class FillModifier(
...
```
JakeWharton: Same `data class` / `@Poko`

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/Size.kt
```diff
@@ -719,6 +758,27 @@ private class WrapContentModifier(
...
+		if (this === other) return true
+		if (other === null) return false
+		if (this::class != other::class) return false
+
+		other as WrapContentModifier
```
JakeWharton, line 731: Why change style from `other !is WrapContentModifier` like the others?

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/Size.kt
```diff
@@ -810,6 +873,13 @@ private class UnspecifiedConstraintsModifier(
 		if (minHeight != Unspecified) minHeight else 0,
 	)
 
+	override fun equals(other: Any?): Boolean {
```
JakeWharton: Same `data class` / `@Poko`

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/layout/PaddingTest.kt
```diff
@@ -0,0 +1,313 @@
...
+		assertThat(Modifier.padding(10, 20))
+			.isEqualTo(Modifier.padding(10, 20, 10, 20))
+	}
+
+	@Test fun paddingLeftNegative() {
+		assertFails {
+			renderMosaic {
+				SingleFiller(Modifier.padding(left = -2))
+			}
+		}
```
JakeWharton: For this and the other invalidate argument tests, the inclusion of the render function and a composable seems redundant.
```suggestion
		assertFails {
			Modifier.padding(left = -2)
		}
```

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/layout/PaddingTest.kt
```diff
@@ -0,0 +1,313 @@
...
+				// Max height
+				assertThat(totalAxisSpacing).isEqualTo(maxIntrinsicHeight(0))
+				assertThat(expectedTotalHeight).isEqualTo(maxIntrinsicHeight(testDimension))
+				assertThat(totalAxisSpacing).isEqualTo(maxIntrinsicHeight(Constraints.Infinity))
+			} catch (t: Throwable) {
+				error = t
+			}
+		}
+
+		error?.let { throw it }
```
JakeWharton: Seems like this behavior should be added directly to `testIntrinsics` helper

JakeWharton (approved):
Amazing stuff!

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/layout/PaddingTest.kt
EpicDima: I looked and checked without the `error` variable at all, it works without it

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/Size.kt
EpicDima: Auto-generated via Android Studio(
Since the others will disappear from the file, I think we can leave this one as it is, while it seems that the implementation is effective.
But the fact is that it is implemented differently in different places of the project and I would like to have one common approach

PR #267, opened 2023-12-02
Introduce Compose ktlint rules

JakeWharton:


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Node.kt
```diff
@@ -14,11 +14,12 @@ import com.jakewharton.mosaic.ui.unit.Constraints
 
 @Composable
 @MosaicComposable
+@Suppress("ktlint") // TODO why doesn't "ktlint:compose:param-order-check" work here?
```
mrmans0n, line 17: Because Modifier doesn't have a default value, as the rule assumes it should have (via another rule). I'll fix this so that if `modifier` doesn't have a default, we don't assume it's going to always be the first one of the optional values in the method signature.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Layout.kt
```diff
@@ -37,15 +37,16 @@ internal sealed class NoContentMeasureScope {
 
 @Composable
 @MosaicComposable
+@Suppress("ktlint:compose:param-order-check") // Order is correct, check just can't tell.
```
mrmans0n, line 40: Yea, it being a `fun interface` messes with the rule here (as it's only looking at lambdas at face value). I might add an extra check in case there is a typealias or a fun interface that's being referenced here, just to cover more edge cases.

JakeWharton: That's what I figured. I wasn't sure if it was possible to check for a `fun interface` or not since there's no type resolution.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Node.kt
JakeWharton: Thanks!

JakeWharton: I will file bugs for these if you don't get to them first.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Layout.kt
mrmans0n: I was thinking on just checking in the current file (so it would at least catch some basic ones), and also allowing to manually add exceptions via configuration (to be able to allowlist types defined elsewhere).

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Node.kt
mrmans0n: I've already added them to my todo list, no worries 😄

PR #270, opened 2023-12-02
Unit tests for Box, Filler, Spacer, Row and Column modifiers

EpicDima:


mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/ui/RowColumnModifierTest.kt
EpicDima, line 1: There are slightly unfair tests here, since it is not the change that is being checked, but two new states. I think for such tests we need something like a test `runMosaic`

JakeWharton:
On vacation this week. Will look at these soon.

JakeWharton (approved):

Issue #3, opened 2020-11-20
Use effects to denote scope of work


JakeWharton:
Currently we build our own coroutine scope to encapsulate all of the work required to render. When this scope ends, rendering ends and the program exits cleanly (modulo internal nonsense to make this work).

Ideally effects inside the composition are used to scope the composition instead. Compose internals currently prevent this, which is tracked as https://issuetracker.google.com/issues/169425431.

JakeWharton:
Today:
```kotlin
fun main() = runMosaic {
  val countValue = mutableStateOf(0)

  setContent {
    val count by remember { countValue }
    Text("The count is: $count")
  }

  for (i in 1..20) {
    delay(250)
    countValue.value = i
  }
}
```
Tomorrow:
```kotlin
fun main() = runMosaic {
  val count by remember { mutableStateOf(0) }
  Text("The count is: $count")

  LaunchedEffect(null) {
    for (i in 1..20) {
      delay(250)
      count++
    }
  }
}
```

JakeWharton:
I was doing some Compose UI for Desktop recently, and it seems like they support this internally. Need to investigate.

m-sasha:
I think what you need is something like `FlushCoroutineDispatcher` from Compose Multiplatform.

JakeWharton:
Done in #284

PR #284, opened 2023-12-19
Change runMosaic to support effects

EpicDima:
An attempt to do https://github.com/JakeWharton/mosaic/issues/3 It works on all samples. 
I haven't checked for js and native(
For jest, it seems that it would be cool to have a `MosaicScope` with a method for exiting, but it's more difficult to do it correctly

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
@@ -12,15 +12,16 @@ import com.github.ajalt.mordant.terminal.Terminal as MordantTerminal
 import com.jakewharton.mosaic.layout.MosaicNode
 import com.jakewharton.mosaic.ui.BoxMeasurePolicy
 import com.jakewharton.mosaic.ui.unit.IntSize
+import kotlin.coroutines.CoroutineContext
 import kotlin.time.ExperimentalTime
-import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.consumeEach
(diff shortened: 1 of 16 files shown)


JakeWharton:
Will have to look when I'm at a computer (on paternity leave until New Year).

Regarding,

> For jest, it seems that it would be cool to have a `MosaicScope` with a method for exiting, but it's more difficult to do it correctly

I think if done properly we shouldn't need a `MosaicScope` anymore. Granted, we may still want one similar to how Compose Desktop provides a scope (I think with exiting functions, too).

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -117,59 +99,32 @@ public suspend fun runMosaic(body: suspend MosaicScope.() -> Unit): Unit = corou
...
-
-	if (hasFrameWaiters) {
-		CompletableDeferred<Unit>().also {
-			displaySignal = it
-			it.await()
-		}
+	val effectJob = checkNotNull(recomposer.effectCoroutineContext[Job]) {
+		"No Job in effectCoroutineContext of recomposer"
 	}
+	effectJob.children.forEach { it.join() }
```
JakeWharton: Compose UI for Desktop does not appear to be doing this for its `application { }` block and `Window`s. Is this not included in the `close()`/`join()` dance below? Granted, they are using subcompositions, so perhaps they just don't anticipate anyone launching effects in the root and have a bug?

JakeWharton (approved):
I'm sorry for sitting on this for so long, especially since I desired this behavior so much early on! Your implementation looks great, and basically mirrors what Compose UI for Desktop is also doing.

If you want to rebase and resolve any conflicts I'm ready to get this in!

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
EpicDima: I looked more closely at the implementations and in general it was possible to preserve the behavior without these instructions, so I removed them.
I don’t remember for sure why they appeared in the first place, but I don’t rule out that updates could have had an impact.

JakeWharton:
Let's go!

PR #298, opened 2024-01-13
Add Binary Compatibility Validator

EpicDima:
Tracking changes to the library's public contract.

https://github.com/Kotlin/binary-compatibility-validator

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,6 +10,7 @@ buildscript {
 		classpath libs.buildconfig.gradlePlugin
 		classpath libs.poko.gradlePlugin
 		classpath libs.spotless.gradlePlugin
+		classpath libs.binary.compatibility.validator.gradlePlugin
 		classpath 'com.jakewharton.mosaic:mosaic-gradle-plugin'
 	}
 	repositories {
@@ -39,17 +40,6 @@ allprojects {
 
 	tasks.withType(KotlinCompile).configureEach {
 		compilerOptions.freeCompilerArgs.add('-Xjvm-default=all')
-	}
-
-	plugins.withId('maven-publish') {
-		publishing {
-			repositories {
-				maven {
-					name = "TestMaven"
-					url = file("${rootProject.buildDir}/testMaven").toURI().toString()
-				}
-			}
-		}
 	}
 
 	apply plugin: 'com.diffplug.spotless'
--- a/publish.gradle
+++ b/publish.gradle
@@ -1,0 +1,11 @@
+apply plugin: "maven-publish"
+apply plugin: "binary-compatibility-validator"
+
+publishing {
+	repositories {
+		maven {
+			name = "TestMaven"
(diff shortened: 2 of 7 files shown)


.github/workflows/build.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
         with:
           gradle-home-cache-cleanup: true
 
-      - run: ./gradlew build :samples:jest:run :mosaic-runtime:dokkaHtml
+      - run: ./gradlew build :samples:jest:run :mosaic-runtime:dokkaHtml apiCheck
```
JakeWharton: This isn't needed, as it should already run as part of `check` which is run as part of `build`

build.gradle
```diff
@@ -70,3 +71,17 @@ allprojects {
 		}
 	}
 }
+
+apply plugin: 'binary-compatibility-validator'
```
JakeWharton: Lets do one of two things:

- Apply it directly to the projects that we want (runtime, Gradle plugin)
- Apply it to all projects which have the publishing plugin applied. The implication being if we are publishing the artifact then we care about the API.

JakeWharton (approved):

.github/workflows/build.yaml
EpicDima: Great, I didn't know there was such a task.

publish.gradle
```diff
@@ -0,0 +1,11 @@
+apply plugin: "maven-publish"
```
EpicDima: Use a short name

publish.gradle
```diff
@@ -0,0 +1,11 @@
...
+apply plugin: "binary-compatibility-validator"
+
+publishing {
+	repositories {
+		maven {
+			name = "TestMaven"
+			url = file("${layout.buildDirectory}/testMaven").toURI().toString()
```
EpicDima: Changed deprecated `buildDir` to an alternative `layout.buildDirectory`

publish.gradle
EpicDima, line 1: I don't know how correct the name of the file in which the API validator is connected is

publish.gradle
JakeWharton: This name is for the Gradle built-in plugin: https://docs.gradle.org/current/userguide/publishing_maven.html

That plugin is applied by https://github.com/vanniktech/gradle-maven-publish-plugin which is the plugin we want to use. The old code keyed off of `maven-publish` because it needed to wait until the `publishing` block was available. Since the application of the plugin is now in the same file, we don't need to wait.
```suggestion
apply plugin: "com.vanniktech.maven.publish"
```

publish.gradle
JakeWharton: Now that it executes in the context of the subproject, we need to use the root projects' build dir. Also skipped a roundtrip through string with some of the `Directory` APIs.
```suggestion
			url = rootProject.layout.buildDirectory.dir("testMaven").get().asFile.toURI().toString()
```

PR #300, opened 2024-01-16
Update compose-rules to 0.3.9 and remove suppressions

mrmans0n:
Updates compose-rules to the latest version, adds the new configuration available to treat `MeasurePolicy` as a lambda (as it is one) and removes all the suppressions. 

Merging this should close #277.

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Layout.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Layout.kt
@@ -37,7 +37,6 @@ internal sealed class NoContentMeasureScope {
 
 @Composable
 @MosaicComposable
-@Suppress("ktlint:compose:param-order-check") // Order is correct, check just can't tell.
 internal fun Layout(
 	modifier: Modifier = Modifier,
 	debugInfo: () -> String = { "Layout()" },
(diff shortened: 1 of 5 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Node.kt
```diff
@@ -14,14 +14,12 @@ import com.jakewharton.mosaic.ui.unit.Constraints
 
 @Composable
 @MosaicComposable
-@Suppress("ktlint") // TODO why doesn't "ktlint:compose:param-order-check" work here?
```
mrmans0n, line 17: It didn't work there because `content` had a default value and it was laid out before other parameters that were non-optional. I did a "Change Signature" IJ refactor here, but let me know if you'd rather have it done as before.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Layout.kt
```diff
@@ -37,16 +37,15 @@ internal sealed class NoContentMeasureScope {
 
 @Composable
 @MosaicComposable
-@Suppress("ktlint:compose:param-order-check") // Order is correct, check just can't tell.
```
mrmans0n, line 40: `NoContentMeasurePolicy` is picked up as a lambda automatically as it is defined in this same file (after 0.3.8 of compose-rules). 

In case this class was moved elsewhere, the spotless config will need to have this class added to the `compose_treat_as_lambda` list -- but as it is now, it's not necessary.

JakeWharton:
Thanks!

PR #337, opened 2024-03-26
Use CLRF line endings to support rendering in raw mode

### LINKED ISSUE #336: Support rendering in raw mode
When entering raw mode in a terminal the automatic translation of `\n` into `\r\n` is disabled, this causes the rendering using mosaic to break since it only outputs `\n` during rendering.

See: https://unix.stackexchange.com/questions/366423/unexpected-indentation-behaviour-when-i-set-the-terminal-to-raw-mode-why-is-th

I suppose not all terminal libraries disable this automatic conversion since the robot sample renders correctly using the jline3 library but many do including https://github.com/crossterm-rs/crossterm which I'm building Kotlin bindings for https://github.com/typfel/crossterm-kmp

I've applied this change in https://github.com/typfel/mosaic/tree/fix/use-crlf-newlines and from my testing it seems to work both cases when raw mode is enabled / disabled.

typfel:
Fixes https://github.com/JakeWharton/mosaic/issues/336

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Changed:
 - Disable klib signature clash checks for JS compilations. These occasionally occur as a result of Compose compiler behavior, and are safe to disable (the first-party JetBrains Compose Gradle plugin also disables them).
 
 Fixed:
-- Use CRLF line endings to fix rendering when a terminal is in raw mode
+- Use CRLF line endings to fix rendering when a terminal is in raw mode.
 
 
 ## [0.11.0] - 2023-02-27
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ Changed:
 - Disable klib signature clash checks for JS compilations. These occasionally occur as a result of Compose compiler behavior, and are safe to disable (the first-party JetBrains Compose Gradle plugin also disables them).
 
 Fixed:
--
+- Use CRLF line endings to fix rendering when a terminal is in raw mode
```
JakeWharton:
```suggestion
- Use CRLF line endings to fix rendering when a terminal is in raw mode.
```

JakeWharton (approved):

PR #357, opened 2024-04-28
Support synchronized update for rendering

### LINKED ISSUE #108: Use terminal sync for rendering
https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036

EpicDima:
I am not 100% sure of the solution, but it seems that it is possible to use this, since those who do not support the terminal feature simply will not react in any way. I'm not sure yet, because I couldn't come up with a sample case that would show the real difference, everything was always fine in both cases.

Closes https://github.com/JakeWharton/mosaic/issues/108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 New:
-- Nothing yet!
+- Support synchronized terminal update for rendering.
 
 Changed:
 - Nothing yet!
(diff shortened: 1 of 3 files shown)


EpicDima:
After creating the pull request, I did an rrtop sample for a while, on it I really saw the difference when synchronized output is enabled and when it is not. So now I'm sure the code is correct 🙂

JakeWharton:
The reason I was hesitant was that I don't like the additional `StringBuilder` that this causes to be allocated. But it's fine for now, and can be improved upon later.

Issue #316, opened 2024-02-22
North star projects


JakeWharton:
I've had this open in a tab for nearly a year so I figured I would move it here.

This is a good example of a north star rendering that we should work towards being able to render.

![](https://raw.githubusercontent.com/wojciech-zurek/rrtop/master/shots/rrtop5.png)

You can see a bunch of components here that we could build today. And it would be interesting to find out what works well and what doesn't.

We can create some simple random sources of data and work towards being able to render this. A great starting point would be the simple bordered boxes with titles, for example.

There's probably other really amazing terminal renderings that people know of which I would also welcome. We don't need to build them all, but it's probably wise to start targeting some concrete eye candy.

turtlekiosk:
To add to the list of pretty TUIs, I really like the https://charm.sh/ projects

JakeWharton:
A bunch here: https://terminaltrove.com/categories/tui/

JakeWharton:
https://toaster.llc/debase/index.html

JakeWharton:
Ink and ratatui have a lot of these built-in to their core. I think because there's only so many ways to rendering things in the terminal we can bake in a few high-level components. Although maybe in a sibling package (akin to what material3 is).

gavr123456789:
- Scala tui: https://github.com/oyvindberg/tui-scala/ 
- Java Turbo Vision reimplementation https://gitlab.com/AutumnMeowMeow/jexer
- Python tui: https://github.com/Textualize/textual
- List of them: https://github.com/dankamongmen/notcurses/blob/master/doc/OTHERS.md

JakeWharton:
https://github.com/aristocratos/btop, which is similar to rtop

JakeWharton:
From #81

![image](https://user-images.githubusercontent.com/2906988/200557515-1a6f91f9-a1d4-46e1-9c57-c0c68d1b0502.png)

Text alignment, border, badass shadow

PR #366, opened 2024-05-13
Add rrtop sample

EpicDima:
An example of a UI like the [rrtop](https://github.com/wojciech-zurek/rrtop) program with **fake** data.

I don't think this is an ideal example from an architectural point of view, but from the point of view of the different capabilities of mosaic at the moment it seems good.

An idea from https://github.com/JakeWharton/mosaic/issues/316

<details>
<summary>A few screenshots (but it's better to try it yourself)</summary>

Screenshot 2024-05-13 at 13 03 37

Screenshot 2024-05-13 at 13 03 45

Screenshot 2024-05-13 at 13 03 59

Screenshot 2024-05-13 at 13 04 14

Screenshot 2024-05-13 at 13 04 28

</details>


JakeWharton:
This is absolutely amazing. I have done a cursory review and it looks great. Going to merge as-is!

PR #372, opened 2024-05-18
Remove autoboxing and object creation from multiple locations

EpicDima:
- Add `Color.Unspecified` for replace nullable `Color`
- Add some extension functions for `Color` to check on `Color.Unspecified`
- Use `TextStyle.Null` instead null for `TextStyle`
- Add some extension functions for `TextStyle` to check on `TextStyle.Null` and `TextStyle.None`
- Make `TextStyle` an inline class
- Using `Int` as codepoint (`TextPixel#codePoint`) instead of `String` in `TextPixel` (`TextPixel#value`)
- Rename `style: TextStyle` to `textStyle: TextStyle` for clearer and less confusing naming, as there are several other styles
- Add emoji to demo sample and use of 2 styles for part of the text
    image

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Color.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Color.kt
@@ -1,3 +1,5 @@
+@file:Suppress("NOTHING_TO_INLINE")
+
 package com.jakewharton.mosaic.ui
 
 import androidx.compose.runtime.Immutable
@@ -7,7 +9,8 @@ import kotlin.jvm.JvmInline
 @Immutable
 @JvmInline
 public value class Color internal constructor(
-	private val value: Int,
+	@PublishedApi
+	internal val value: Int,
 ) {
 	override fun toString(): String = "Color($redInt, $greenInt, $blueInt)"
 
@@ -85,6 +88,9 @@ public value class Color internal constructor(
 
 	public companion object {
 		@Stable
+		public val Unspecified: Color = Color(UnspecifiedColor)
+
+		@Stable
 		public val Black: Color = Color(0x000000)
 
 		@Stable
@@ -109,6 +115,9 @@ public value class Color internal constructor(
 		public val White: Color = Color(0xFFFFFF)
 	}
 }
+
+@PublishedApi
+internal const val UnspecifiedColor: Int = Int.MIN_VALUE
 
 /**
  * Creates a new [Color] instance from an RGB color components.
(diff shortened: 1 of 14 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/TextStyle.kt
```diff
@@ -1,35 +1,80 @@
...
-	public operator fun contains(style: TextStyle): Boolean = (style.bits and bits) != 0
+	public operator fun contains(other: TextStyle): Boolean = (bits and other.bits) == other.bits
 
 	public companion object {
 		@Stable
-		public val None: TextStyle = TextStyle(0)
+		public val Null: TextStyle = TextStyle(NullTextStyle)
+
+		@Stable
+		public val None: TextStyle = TextStyle(NoneTextStyle)
```
JakeWharton: What are the semantics here? Null resets to no style and none is a no-op? Or the opposite maybe? I think we need better names here, and definitely some docs.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Color.kt
```diff
@@ -110,6 +116,9 @@ public value class Color internal constructor(
 	}
 }
 
+@PublishedApi
+internal const val UnspecifiedColor: Int = Int.MIN_VALUE
```
JakeWharton, line 120: Are we just assuming no one is going to use 0x80000000? Or do we basically ignore alpha channel? I don't recall after the switch to true color seeing whether it was supported or not.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/TextStyle.kt
```diff
@@ -1,35 +1,132 @@
...
+		 * since it does not replace the style that was previously specified.
+		 *
+		 * In the example below, we will get `789456` in the console, where everything will be in bold,
+		 * as indicated in the first [Text]. The second [Text] changes `123` to `789`,
+		 * but does not change the previously specified style in any way.
+		 *
+		 * ```kotlin
+		 * Box {
+		 * 	Text("123456", textStyle = TextStyle.Invert)
+		 * 	Text("789", textStyle = TextStyle.Null)
```
JakeWharton:
```suggestion
		 * 	Text("789", textStyle = TextStyle.Unspecified)
```

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/TextStyle.kt
```diff
@@ -1,35 +1,132 @@
...
+		 *
+		 * ```kotlin
+		 * Box {
+		 * 	Text("123456", textStyle = TextStyle.Bold)
+		 * 	Text("789", textStyle = TextStyle.Null)
+		 * }
+		 * ```
+		 *
+		 * Output to the console:
+		 * 789**456**
```
JakeWharton: Output doesn't match. Null needs updated too.

JakeWharton (approved):

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/Color.kt
EpicDima: Yes, I specifically chose a well-known constant for simplicity, and yes, now we ignore the alpha channel, since, at least, I did not see the possibility of using it in the terminal. If we want an alpha channel, then we just need to change the field in Color to long, because otherwise 32 bits will not be enough for such a constant.

JakeWharton: The only way we'd support alpha is to premultiply it. So if you had a background of red and chose yellow at 50% we'd do the math and turn the text opaque orange. Not with the bother at this time.

JakeWharton:
Great stuff. Was mobile-only for a week which made it hard to review properly.

PR #383, opened 2024-05-29
Modify `DrawScope#drawRect`

EpicDima:
- Change the `drawRect` method in `DrawScope` like the method from the Compose UI
- Add support for `DrawStyle`: `Fill` and `Stroke` in `DrawScope#drawRect`
- Add support for drawing a rectangle from text characters
- Add `Filler` with code point
- Change the robot sample to show `drawRect` with `Stroke`:
  Screenshot 2024-05-29 at 22 22 48

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawStyle.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawStyle.kt
@@ -1,0 +1,28 @@
+package com.jakewharton.mosaic.layout
+
+import dev.drewhamilton.poko.Poko
+
+/**
+ * Represents how the shapes should be drawn within a [DrawScope]
+ */
+public sealed interface DrawStyle
+
+/**
+ * Default [DrawStyle] indicating shapes should be drawn completely filled in with the
+ * provided color or pattern
+ */
+public data object Fill : DrawStyle
+
+/**
+ * [DrawStyle] that provides information for drawing content with a stroke
+ *
+ * @param width Configure the width of the stroke in cells
+ */
+@Poko
+public class Stroke(
+	internal val width: Int = 1,
+) : DrawStyle {
+	override fun toString(): String {
+		return "Stroke(width=$width)"
+	}
+}
--- a/samples/robot/src/main/kotlin/example/robot.kt
+++ b/samples/robot/src/main/kotlin/example/robot.kt
@@ -3,8 +3,11 @@ package example
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.setValue
+import com.jakewharton.mosaic.layout.Stroke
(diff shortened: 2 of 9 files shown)


samples/robot/src/main/kotlin/example/robot.kt
```diff
@@ -31,10 +37,16 @@ fun main() = runMosaicBlocking {
...
+			Text("Position: $x, $y   Robot: $robotWidth, $robotHeight   World: $worldWidth, $worldHeight")
 			Spacer(Modifier.height(1))
 			// TODO https://github.com/JakeWharton/mosaic/issues/11
-			Box(modifier = Modifier.size(width, height).offset { IntOffset(x, y) }) {
+			Box(
+				modifier = Modifier
+					.drawBehind { drawRect(worldBorderChar, drawStyle = Stroke(worldBorderWidth)) }
+					.padding(worldBorderWidth)
+					.size(worldWidth, worldHeight)
+					.offset { IntOffset(x, y) },
```
JakeWharton, line 48: Seems like offset should be on the `Text` rather than the `Box`

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawStyle.kt
```diff
@@ -0,0 +1,28 @@
...
+ * [DrawStyle] that provides information for drawing content with a stroke
+ *
+ * @param width Configure the width of the stroke in cells
+ */
+@Poko
+public class Stroke(
+	internal val width: Int = 1,
+) : DrawStyle {
+	override fun toString(): String {
+		return "Stroke(width=$width)"
```
JakeWharton, line 26: Isn't this the default `toString()` that Poko would generate?

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawStyle.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Represents how the shapes should be drawn within a [DrawScope]
+ */
+public sealed interface DrawStyle
+
+/**
+ * Default [DrawStyle] indicating shapes should be drawn completely filled in with the
+ * provided color or pattern
+ */
+public data object Fill : DrawStyle
```
JakeWharton, line 14: I would think these types would be nested inside `DrawStyle` since it's enum-like.

JakeWharton (approved):

JakeWharton:
I haven't forgotten about the other PRs. Still catching up from KotlinConf...

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/DrawStyle.kt
EpicDima: I think so too, but the idea was for Mosaic to be more similar in this regard to the Compose UI, where it is done this way ¯\\\_(ツ)_/¯

PR #394, opened 2024-06-08
Reusing canvas for rendering

EpicDima:
Partial solution for https://github.com/JakeWharton/mosaic/issues/153.

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/canvas.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/canvas.kt
@@ -1,3 +1,5 @@
+@file:Suppress("NOTHING_TO_INLINE")
+
 package com.jakewharton.mosaic
 
 import com.jakewharton.mosaic.ui.AnsiLevel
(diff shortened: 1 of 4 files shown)


mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/AnsiRenderingTest.kt
```diff
@@ -178,11 +178,11 @@ class AnsiRenderingTest {
 
 		assertThat(rendering.render(rootNode).toString()).isEqualTo(
 			"""
-			|One
-			|Two
+			|One $s
+			|Two $s
```
JakeWharton: This seems worse. I don't think we need to optimize statics to reuse the same surface since they're rare, and even when used there's probably only one.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/canvas.kt
```diff
@@ -26,16 +28,68 @@ internal interface TextCanvas {
...
+		private set
+
 	override var translationX = 0
 	override var translationY = 0
 
-	private val rows = Array(height) { Array(width) { TextPixel(' ') } }
+	private var fullWidth: Int = width
+	private var fullHeight: Int = height
+
+	private val rows = ArrayList(fullHeight) { ArrayList(fullWidth) { TextPixel(' ') } }
```
JakeWharton: I don't see a big motivation from switching from `Array` here. Is it because of row resizing? We can accomplish that with a helper https://pl.kotl.in/c_ED2lc0h. I'll file a Kotlin issue to provide such an overload.

JakeWharton: https://youtrack.jetbrains.com/issue/KT-70984

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/AnsiRenderingTest.kt
EpicDima: Yes, I know it looks worse here. I did this consciously, because I made https://github.com/JakeWharton/mosaic/pull/439 where I delete empty characters at the end of lines, which, in my opinion, eliminates this problem later.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/canvas.kt
EpicDima: Yes, I changed it only to change the size out of the box.

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/AnsiRenderingTest.kt
JakeWharton: I think trimming at the end makes sense for an individual surface, especially the main one. I'll get to that PR soon.

But by combining the static surfaces we're now doing more work to merge them and allocate more space, only to then have to process more to trim the new trailing spaces. And ultimately it results in no change in the output but more code to maintain.

Since in practice there aren't multiple static nodes, and even if there are they're all not rendering at the same time, and even if multiple render they're not rendering much, doing this still doesn't gain us anything.

JakeWharton:
Are you okay if I close this? Or do you want to push back more on my claim that this is an uncommon problem?

EpicDima:
No problem if you don't think it's necessary 🙂

PR #395, opened 2024-06-08
Remove background color from default `rrtop` sample theme

### LINKED ISSUE #390: rrtop theme with no explicit background
The explicit background makes the recordings render poorly

EpicDima:
Closes https://github.com/JakeWharton/mosaic/issues/390.


samples/rrtop/src/main/kotlin/example/palletes/DefaultColorsPalette.kt
```diff
@@ -3,7 +3,7 @@ package example.palletes
 import com.jakewharton.mosaic.ui.Color
 import example.RrtopColorsPalette
 
-private val Black = Color(0, 0, 0)
+private val NoColor = Color.Unspecified
```
JakeWharton: Inline this?

JakeWharton (approved):

Issue #4, opened 2020-11-20
Testing support


JakeWharton:
The counter example needs tests to show off how they work or would work.

One should start the rendering and assert 0, delay 250, and assert 1. The other should start the rendering and fast-forward to 20 and assert it exits cleanly and does not render anything else.

JakeWharton:
Generalizing this to simply be a testing framework for all programs built on Mosaic

JakeWharton:
Been thinking about this a bit and bouncing ideas off of folks.

I think we want two APIs:
 - A `Turbine`-like harness which allows pulling frames which (by default) omit all ANSI escapes and instead return text-only "frames".
   - You could optionally turn on ANSI escapes, although it'll be very noisy.
   - You can send events like input and resize
 - A Paparazzi-like harness which allows capturing images of frames and comparing golden images against the running program.
   - Need some kind of software-based renderer to avoid platform-specific rendering problems. Maybe libghostty once it's public, and presuming it includes an actual software renderer.

JakeWharton:
Another feature that might be interesting is rendering spaces with a middot (&middot;). Probably better to fix rendering so we omit trailing spaces altogether though...

PR #397, opened 2024-06-08
Add unit test for counter

EpicDima:
The first step for https://github.com/JakeWharton/mosaic/issues/4

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
@@ -52,7 +52,14 @@ public fun renderMosaic(content: @Composable () -> Unit): String {
 	return render.toString()
 }
 
-public suspend fun runMosaic(content: @Composable () -> Unit): Unit = coroutineScope {
+public suspend fun runMosaic(content: @Composable () -> Unit) {
+	runMosaic(content, ::platformDisplay)
+}
+
+internal suspend fun runMosaic(
+	content: @Composable () -> Unit,
+	display: (CharSequence) -> Unit,
+): Unit = coroutineScope {
 	val terminal = MordantTerminal()
 
 	val rendering = if (debugOutput) {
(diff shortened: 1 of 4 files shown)


mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/MosaicTest.kt
```diff
@@ -24,4 +29,51 @@ class MosaicTest {
 			""".trimMargin().replaceLineEndingsWithCRLF(),
 		)
 	}
+
+	@Test fun counter() {
+		runTest {
```
JakeWharton:
```suggestion
	@Test fun counter() = runTest {
```
`runTest` returns a `Promise<Unit>` on JS that the testing harness will `await()`. Otherwise the test won't actually run (or, at least, won't run past the first suspension point).

JakeWharton (approved):

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/MosaicTest.kt
EpicDima: Probably the first time I wrote `runTest` without `return` and overlooked this behavior, thank you!

EpicDima:
The checks failed without logs and everything was successful locally, I don't know what the problem is(

EpicDima:
> The checks failed without logs and everything was successful locally, I don't know what the problem is(

I think there was a problem with CI and running jobs

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/MosaicTest.kt
```diff
@@ -24,4 +29,49 @@ class MosaicTest {
...
+			},
+		)
+
+		assertThat(strings.size).isEqualTo(22)
+		assertThat(strings[0]).isEqualTo(
+			"""
+			|${ansiBeginSynchronizedUpdate}The count is: 0
+			|$ansiEndSynchronizedUpdate
+			""".trimMargin().replaceLineEndingsWithCRLF(),
+		)
```
JakeWharton, line 62: I added some thoughts at https://github.com/JakeWharton/mosaic/issues/4#issuecomment-2285275775. I don't want to jump the gun on too many of these tests without building a good abstraction for them. We have a pretty good track record of building Turbine-like APIs for this and have it be successful.

Here's one example: https://github.com/cashapp/redwood/blob/831b4c9d7ced73324c6bd8098e770900b9158c39/samples/counter/presenter/src/commonTest/kotlin/com/example/redwood/counter/presenter/CounterTest.kt#L30-L50

EpicDima: I completely agree that it is necessary to make an abstraction, I just wanted there to be a minimal simple unit test from which to build further.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -106,16 +113,22 @@ public suspend fun runMosaic(content: @Composable () -> Unit): Unit = coroutineS
...
 			CompositionLocalProvider(LocalTerminal provides terminalInfo.value) {
 				content()
 			}
 		}
 
+		val effectJob = checkNotNull(recomposer.effectCoroutineContext[Job]) {
+			"No Job in effectCoroutineContext of recomposer"
+		}
+		effectJob.children.forEach { it.join() }
+		recomposer.awaitIdle()
```
EpicDima, line 130: It turned out to be necessary to pass the `counter` unit test, without this, the last render is not done with `TestDispatcher`s. If you specify `Dispatchers.Default` or make a `delay` after assigning a value to `count`, it will work, but it doesn't seem like it's super.

PR #402, opened 2024-06-10
Its own platform implementation of different things

EpicDima:
This is a draft, then you need to add user input processing and change the samples


JakeWharton:
Let's do Mordant's new support for now, but I'm still open to doing this ourselves. Especially since there's a bunch of cool stuff in modern terminals that would be fun to expose which Mordant does not handle.

EpicDima:
Mordant does not support user input for JS, so I decided to hold off on integrating user input into Mosaic.

I think this pull request can be closed. I hope it will be on the closed list if there is a need to look at it.

JakeWharton:
I'm going to expect/actual it. For now in my branch I just dropped JS support, though.

PR #433, opened 2024-08-17
Create internal `Turbine`-like abstraction for `Mosaic` unit testing

EpicDima:
---

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
@@ -5,6 +5,7 @@ import androidx.compose.runtime.BroadcastFrameClock
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Composition
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.Recomposer
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.snapshots.Snapshot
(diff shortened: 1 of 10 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -53,86 +54,112 @@ public fun renderMosaic(content: @Composable () -> Unit): String {
...
+				delay(50L)
+			}
 		}
 	}
+}
 
-	val recomposer = Recomposer(composeContext)
-	launch(composeContext, start = CoroutineStart.UNDISPATCHED) {
-		recomposer.runRecomposeAndApplyChanges()
+internal abstract class BaseMosaicComposition(coroutineScope: CoroutineScope) {
```
JakeWharton, line 103: Would prefer this be called `MosaicComposition` and the subtypes be named with something else like `TestMosaicComposition` and `RenderingMosaicComposition` (or something)

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/rendering.kt
```diff
@@ -81,6 +81,8 @@ internal class AnsiRendering(
 		return stringBuilder.apply {
 			clear()
 
+			append(ansiBeginSynchronizedUpdate)
```
JakeWharton, line 84: So glad these are here now. It pained me to do all this `StringBuilder` stuff and then have a second `StringBuilder` just to wrap these around it.

At some point I want to experiment with switching the whole pipeline to deal in bytes rather than chars/Strings.

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/TestMosaicComposition.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.jakewharton.mosaic.ui.AnsiLevel
+import com.jakewharton.mosaic.ui.unit.IntSize
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.withTimeout
+
+private val DefaultTestTerminalSize = IntSize(80, 20)
+
+internal suspend fun runMosaicTest(
```
JakeWharton, line 14: Up next:
- Move this to mosaic-testing module
- Extract an interface for the control APIs + `setContent`.

JakeWharton (approved):

PR #434, opened 2024-08-18
Unit tests for Row and Column

EpicDima:
And fix unit tests after merging two implicitly conflicting pull requests.

---

### DIFF
--- a/mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/ui/RowColumnTest.kt
+++ b/mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/ui/RowColumnTest.kt
@@ -1,0 +1,2399 @@
+package com.jakewharton.mosaic.ui
+
+import androidx.compose.runtime.Composable
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import assertk.assertions.isZero
+import com.jakewharton.mosaic.ConstrainedBox
+import com.jakewharton.mosaic.Container
+import com.jakewharton.mosaic.Holder
+import com.jakewharton.mosaic.layout.aspectRatio
+import com.jakewharton.mosaic.layout.fillMaxHeight
+import com.jakewharton.mosaic.layout.fillMaxWidth
+import com.jakewharton.mosaic.layout.height
+import com.jakewharton.mosaic.layout.heightIn
+import com.jakewharton.mosaic.layout.padding
+import com.jakewharton.mosaic.layout.requiredSize
+import com.jakewharton.mosaic.layout.size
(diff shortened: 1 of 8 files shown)


mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/ui/RowColumnTest.kt
```diff
@@ -0,0 +1,2399 @@
...
+
+		val gap = (rootNode.height - size * 3) / 3
+
+		val firstChildContainerNode = rootNode.children[0].children[0].children[0]
+		val secondChildContainerNode = rootNode.children[0].children[0].children[1]
+		val thirdChildContainerNode = rootNode.children[0].children[0].children[2]
+
+		assertThat(firstChildContainerNode.position).isEqualTo(IntOffset(0, gap / 2))
+		assertThat(secondChildContainerNode.position).isEqualTo(IntOffset(0, size + (gap * 3 / 2)))
+		assertThat(thirdChildContainerNode.position).isEqualTo(IntOffset(0, size * 2 + (gap * 5 / 2)))
```
JakeWharton, line 1519: I'm not sure whether it makes sense to do math in all these tests or to put hard coded values here (since the input values are hard coded). Since you did them all with math we can leave them for now, but I _think_ I would prefer hard coded values in the future.

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/ui/RowColumnTest.kt
```diff
@@ -0,0 +1,2399 @@
...
+
+		assertThat(firstChildContainerNode.size).isEqualTo(IntSize(rootWidth, height))
+		assertThat(firstChildContainerNode.position).isEqualTo(IntOffset.Zero)
+		assertThat(secondChildContainerNode.size).isEqualTo(IntSize(0, height))
+		assertThat(secondChildContainerNode.position).isEqualTo(IntOffset(rootWidth, 0))
+	}
+
+	@Test fun testRow_invalidWeight() {
+		with(RowScopeInstance) {
+			assertFails {
```
JakeWharton, line 164: Would like to see at minimum asserting on the exception type here, if not also the message.

JakeWharton (approved):
I mostly just skimmed these. There's too much to review, and I trust you enough now to have covered the behavior adequately.

EpicDima:
These are unit tests from the Compose Foundation adapted for Mosaic, I think they work correctly. 
I will fix other review issues in the next pull requests.

JakeWharton:
If they use the math approach for asserting the results then we can keep doing that so they mostly in sync.

PR #435, opened 2024-08-19
Use `runMosaicTest` for all unit tests

EpicDima:
- Rename `BaseMosaicComposition` to `MosaicComposition`, and `MosaicComposition` to `RenderingMosaicComposition` (with some changes)
- Use `RenderingMosaicComposition` in `renderMosaic` instead of `mosaicNodes`
- Use `runMosaicTest` in unit tests instead of `mosaicNodes`
- Remove `mosaicNodes`
- Create an `internal` `TestMosaicComposition` interface with the implementation of `RealTestMosaicComposition`
- Add unit tests for synchronous `renderMosaic`

---
The main number of changes is simply the transfer of tests to `runMosaicTest`.

---

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
@@ -11,9 +11,13 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.snapshots.Snapshot
 import com.github.ajalt.mordant.terminal.Terminal as MordantTerminal
 import com.jakewharton.mosaic.layout.MosaicNode
+import com.jakewharton.mosaic.ui.AnsiLevel
 import com.jakewharton.mosaic.ui.BoxMeasurePolicy
 import com.jakewharton.mosaic.ui.unit.IntSize
 import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
 import kotlin.time.ExperimentalTime
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart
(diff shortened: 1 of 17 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -30,50 +33,45 @@ import kotlinx.coroutines.launch
...
+			coroutineScope = this,
+			onRender = ::platformDisplay,
+		)
 		mosaicComposition.setContent(content)
 		mosaicComposition.awaitComplete()
 	}
 }
 
-internal class MosaicComposition(
+public suspend fun renderMosaic(content: @Composable () -> Unit): String {
```
JakeWharton: Since the initial composition is synchronous, the idea was that this function could always be synchronous, too, and that no effects would run at all. I would strongly prefer to keep a design where we can retain this behavior. I believe we can create our own `CoroutineScope`, rely on `onRender` being called synchronously, and then cancel everything to keep it that way.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -100,16 +98,14 @@ internal class MosaicComposition(
...
+internal abstract class MosaicComposition(coroutineScope: CoroutineScope) {
+	protected abstract val onEndChanges: (MosaicNode) -> Unit
 
 	private val job = Job(coroutineScope.coroutineContext[Job])
 	private val clock = BroadcastFrameClock()
 	protected val composeContext: CoroutineContext = coroutineScope.coroutineContext + job + clock
 
-	private val rootNode = createRootNode()
-	private val applier = MosaicNodeApplier(rootNode) { onRender(rendering.render(rootNode)) }
+	private val applier = MosaicNodeApplier(createRootNode()) { rootNode -> onEndChanges(rootNode) }
```
JakeWharton, line 135:
```suggestion
	private val applier = MosaicNodeApplier(createRootNode(), onEndChanges)
```

EpicDima: I also wanted to do this, but then Android Studio writes such a warning
image

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -30,50 +33,51 @@ import kotlinx.coroutines.launch
...
+			if (singleRender == null) {
+				singleRender = render.toString()
+			}
+		},
+	)
+	mosaicComposition.setContent(content)
+	mosaicComposition.cancel()
+	return checkNotNull(singleRender) {
+		"It shouldn't happen. We are waiting for synchronous render after calling setContent."
+	}
```
JakeWharton:
```suggestion
	return singleRender!!
```
No need to waste bytes on branches that should be impossible

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -30,50 +33,51 @@ import kotlinx.coroutines.launch
...
 
-internal class MosaicComposition(
+public fun renderMosaic(content: @Composable () -> Unit): String {
+	var singleRender: String? = null
+	val mosaicComposition = RenderingMosaicComposition(
+		coroutineScope = CoroutineScope(EmptyCoroutineContext),
+		onRender = { render ->
+			if (singleRender == null) {
+				singleRender = render.toString()
+			}
```
JakeWharton:
```suggestion
			if (singleRender == null) {
				singleRender = render.toString()
			} else {
				throw AssertionError()
			}
```

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -219,3 +227,22 @@ internal class GlobalSnapshotManager {
 		}
 	}
 }
+
+// TestMosaicComposition is located here, not in the commonTest, because task
+// ':mosaic-runtime:compileTestKotlinLinuxArm64' fails with the message:
+// Refined declaration 'fun setContent(content: @Composable() ComposableFunction0<Unit>): Unit'
+// overrides declarations with different or no refinement from: interface TestMosaicComposition : Any
```
JakeWharton, line 254: We should isolate and file a Kotlin bug

JakeWharton: Plus this needs to move to 'mosaic-testing' eventually, probably with some fancy mechanism so our tests can still expose the nodes.

JakeWharton (approved):
What's the primary motivation to remove `mosaicNodes`? I think the `AnsiRenderingTest` and `DebugRenderingTest` have gotten worse with this change. The other tests which operate at a higher-level are fine, but those two now have to do a bunch of boilerplate to test a lower-level component through a higher-level API.

I think we can can have an `internal fun renderMosaicNode` function that is 99% of the existing `renderMosaic` but which returns a `MosaicNode`. Then `renderMosaic` _should_ just be `return renderMosaicNode(content).toString()`. And that allows keeping the two renderer tests a bit simpler.

JakeWharton:
Just did a little diff minimization. I have some other suggestions/changes but I don't want to block this any further and I'll probably do them myself tonight.

PR #442, opened 2024-08-21
Add rudimentary key handling

### LINKED ISSUE #10: Add jline3 and expose raw keyboard input somehow

JakeWharton:
This forces all usages into raw mode, regardless of whether you are listening for keys or not.

Closes #10 

---


JakeWharton:
When `rrtop` sample exits I can no longer type 😑

JakeWharton:
Seems like new snapshot of Mordant fixes that, but breaks key reading altogether.

samples/rrtop/src/main/kotlin/example/RrtopApp.kt
```diff
@@ -39,7 +34,21 @@ fun RrtopApp(rrtopViewModel: RrtopViewModel, colorsPalette: RrtopColorsPalette)
 			modifier = Modifier
 				.width(terminal.size.width)
 				.height(terminal.size.height - 1) // subtraction of one is necessary, because there is a line with a cursor at the bottom, which moves up all the content
-				.background(LocalRrtopColorsPalette.current.mainBg),
+				.background(LocalRrtopColorsPalette.current.mainBg)
+				.onKeyEvent {
+					when (it) {
+						KeyEvent("ArrowLeft") -> rrtopViewModel.onArrowLeftPress()
+						KeyEvent("ArrowRight") -> rrtopViewModel.onArrowLeftPress()
```
EpicDima:
```suggestion
						KeyEvent("ArrowRight") -> rrtopViewModel.onArrowRightPress()
```

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/layout/KeyModifier.kt
```diff
@@ -0,0 +1,65 @@
...
+	 * This function is called when a [KeyEvent] is received by this node during the upward pass.
+	 * While implementing this callback, return true to stop propagation of this event. If you
+	 * return false, the key event will be sent to this [KeyModifier]'s parent.
+	 */
+	public fun onKeyEvent(event: KeyEvent): Boolean
+}
+
+@[Immutable Poko]
+public class KeyEvent(
+	public val key: String,
```
EpicDima, line 27: Is there a list of the names of all the keys somewhere?
I think it would be convenient to have a set of frequently used keys inside the library for quick and convenient use by client applications.

JakeWharton: This is not the final API. I don't like this API at all, actually. It's a direct mapping from Mordant, and that's enough to unblock users who need key input (there's at least one).

Long term a better API that has actual constants for keys and supports the full range of flags should be designed.

ajalt: This is off topic for this PR, but if you don't like the mordant key event API, I'd love to hear your feedback so that I can improve it.

JakeWharton: Yeah I don't mean this necessarily as a criticism of your design choices, more that I'm not sure it makes sense for a relatively high-level UI abstraction like our Compose UI-based composables. We'll probably model ourselves _very_ close to what Compose UI is already doing just for the sake of familiarity.

I'm also just a little wary using strings as the domain of possible values as opposed to some closed set of constants. I know it's a bit tricky at your layer of abstraction because you receive readable text input in addition to key presses which can include multi-codepoint grapheme clusters like 🧑‍🌾.

samples/rrtop/src/main/kotlin/example/RrtopApp.kt
JakeWharton: Nice catch! Applied locally.

JakeWharton:
Still need:
- [x] Test native key event handling. Should probably port another sample to MPP like robot.
- [x] Mordant stable release

Otherwise seems all working!

PR #249, opened 2023-11-13
Initialize AnsiConsole before usage

TWiStErRob:
Fixes #80

PR #443, opened 2024-08-21
Using print for output

EpicDima:
I tried to compare the option before and the option after, I didn't really see the difference, maybe there are cases when it is noticeable? I'll be glad if you tell me.

I wanted to make an issue with the question first, but decided that it would be faster, because there are minimal changes and immediately, if everything is good, we can apply them.

---

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -21,7 +21,6 @@ spotless-gradlePlugin = "com.diffplug.spotless:spotless-plugin-gradle:6.25.0"
 ktlint-core = "com.pinterest.ktlint:ktlint-cli:1.3.1"
 ktlint-composeRules = "io.nlopez.compose.rules:ktlint:0.4.10"
 
-jansi = "org.fusesource.jansi:jansi:2.4.1"
 mordant = "com.github.ajalt.mordant:mordant:2.7.2"
 codepoints = "de.cketti.unicode:kotlin-codepoints:0.9.0"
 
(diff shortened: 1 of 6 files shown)


JakeWharton:
I believe it was for Windows users

JakeWharton:
#249

JakeWharton:
Maybe using Mordant's `Terminal`'s print solves Windows? I don't have a Windows machine to test anything.

EpicDima:
I checked several options on not the newest Windows 10 (x86, latest JDK17) with a built-in terminal (the same with PowerShell) and with the latest version [Terminal](https://github.com/microsoft/terminal) (there is a regular terminal inside, but you can choose others if installed). I tested it on a _jest_ sample. I didn't see any problems with the output other than what is below.

<details>
<summary>1. no changes in simple terminal</summary>

jest_default_simple
</details>

<details>
<summary>2. no changes in Terminal</summary>

jest_default
</details>

<details>
<summary>3. using `print` in simple terminal</summary>

jest_only_print_simple
</details>

<details>
<summary>4. using `print` in Terminal</summary>

jest_only_print
</details>

<details>
<summary>5. using `System.out.write` in Terminal</summary>

jest_system_out
</details>

Others do not differ from the above.

EpicDima:
I will close the pull request, since the changes break the output and it turns out to be incorrect.

JakeWharton:
We should try with Mordant and with raw mode enabled, too. Or maybe the amount of code to make it work is small enough to take on ourselves (or maybe Mordant would take it on).

EpicDima:
I checked with Mordant `Terminal#rawPrint`, the results are one-to-one, as with `print`.

JakeWharton:
Tracking at #471

Issue #190, opened 2023-06-11
Allocation reduction pass


JakeWharton:
Just a reminder to do a normal allocation-reduction pass for free performance wins, even if minute.

- Switch loops to be indexed when we can trust the source of the list to be random access.
- Replace multi-step collection manipulation with traditional loops.
- Reduce/eliminate points of value class or JVM primitive boxing, if possible
- Try to use specialized androidx.collections

etc.

EpicDima:
I wanted to ask, are we doing optimizations right now, while there probably aren't any critical performance problems, and the actual question is, isn't this premature optimization?
If not, why not?
I like this, but it's interesting to get an opinion on this 😊

JakeWharton:
Yeah it's a good question. I have a few opinions on it:

- Speed is a feature.

  Computers felt faster 20 years ago then they do today because abstraction and indirection has outpaced all variants of Moore's law (clock speed, core count, cpu cache size, cpu cache speed, etc.). Fast things _feel_ good, and I want this to be as fast as possible.

- We are the inner loop of applications.

  Someone using an iterator-based loop instead of an indexed-based loop in a Mosaic application is probably not a big deal. Us using an iterator-based loop instead of an indexed-based loop in code that runs on every node occurs multiple orders of magnitude more, which means it has multiple orders of magnitude greater.

  Micro optimizations reducing CPU, memory, or I/O within the library become a macro optimization because we do it at 1000 FPS or within 1000 nodes (or both!).

- It's not that hard!

  Kotlin gives us plenty of tools to allow efficient things to remain expressive. It's not a huge sacrifice as opposed to some other languages. We aren't losing guardrails, we're becoming sympathetic how the machine works.

I've had a lot of success doing this in the past, and generally try to do it in libraries everywhere.

PR #451, opened 2024-08-25
Launch different phases separately depending on need

### LINKED ISSUE #449: Support running Layout and Drawing phases separately
Now the _Composition_ phase is the main one for rendering the UI, since after calling the `onEndChanges` method in the `Applier`, we run the _Layout_ and _Drawing_ phases, respectively. 
Unfortunately, now we do not trigger the _Layout_ or _Drawing_ phases separately, respectively, we cannot optimize some things, the simplest example is `Modifier.offset {}` or `Modifier.drawBehind {}`.
Below I made a simple sample that clearly shows that this behavior is not supported now. This can be circumvented in most cases by dividing the `Composable` function into separate granular ones and using other `Modifier`s, if any.
It seems like it would be cool to natively support the ability to trigger different phases separately, so that you don't have to do workaround on the client code side.

<details>
<summary>Sample</summary>

```kotlin
import androidx.compose.runtime.Composable
import androidx.compose.runtime.LaunchedEffect
import androidx.compose.runtime.getValue
import androidx.compose.runtime.mutableIntStateOf
import androidx.compose.runtime.mutableStateOf
import androidx.compose.runtime.remember
import androidx.compose.runtime.setValue
import com.jakewharton.mosaic.layout.background
import com.jakewharton.mosaic.layout.drawBehind
import com.jakewharton.mosaic.layout.offset
import com.jakewharton.mosaic.layout.width
import com.jakewharton.mosaic.modifier.Modifier
import com.jakewharton.mosaic.runMosaicBlocking
import com.jakewharton.mosaic.ui.Color
import com.jakewharton.mosaic.ui.Column
import com.jakewharton.mosaic.ui.Text
import com.jakewharton.mosaic.ui.unit.IntOffset
import kotlinx.coroutines.delay
import kotlin.jvm.JvmName

@Composable
fun WorkingSample() {
    var color by remember { mutableStateOf(Color.Red) }
    var offsetX by remember { mutableIntStateOf(0) }

    Text(
        value = "X",
        modifier = Modifier
            .background(color)
            .offset(offsetX, 0)
    )

    LaunchedEffect(Unit) {
        delay(500L)
        color = Color.Blue
        offsetX = 5
        delay(1000L)
    }
}

@Composable
fun NotWorkingSample() {
    var color by remember { mutableStateOf(Color.Red) }
    var offsetX by remember { mutableIntStateOf(0) }

    Text(
        value = "X",
        modifier = Modifier
            .drawBehind { drawRect(background = color) }
            .offset { IntOffset(offsetX, 0) }
    )

    LaunchedEffect(Unit) {
        delay(1000L)
        color = Color.Blue
        offsetX = 5
        delay(1000L)
    }
}

fun main() = runMosaicBlocking {
    Column(modifier = Modifier.background(Color.Black).width(10)) {
        WorkingSample()
        NotWorkingSample()
    }
}
```
</details>

<details>
<summary>Screenshot of the output</summary>

The first render:
image

After completing the program:
image
</details>

EpicDima:
A simple implementation that remembers the readings of the states when calling the layout, drawing blocks separately, and then with the help of a global observer checks which states have changed, if those that were read during layout or drawing have changed and composition is not going on now, then we run the layout or drawing phase separately.

Since composition does not occur in case of some changes, therefore, when completed in the `MosaicComposition#awaitComplete` method, we additionally expect one frame to be rendered, if necessary.

Fixes https://github.com/JakeWharton/mosaic/issues/449.

---


JakeWharton:
Will look soon. Been so busy I can barely work on anything myself let alone review.

JakeWharton:
Is this derived from what Compose UI is doing?

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -155,21 +158,82 @@ internal class MosaicComposition(
...
+
+	private fun doLayout(rootNode: MosaicNode) {
+		needLayout = false
+		Snapshot.observe(readObserver = layoutBlockStateReadObserver) {
+			rootNode.measureAndPlace()
+		}
+		doDrawing(rootNode)
+	}
+
+	private fun doDrawing(rootNode: MosaicNode) {
```
JakeWharton: Here and everywhere else, I would prefer "draw" and not "drawing" just to agree more with "layout" as a noun.

Also maybe "perform" instead of "do". There's nothing technically wrong with "do", it's just kind of weak as far as verbs go.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -155,21 +158,82 @@ internal class MosaicComposition(
...
-	private val applier = MosaicNodeApplier(onEndChanges)
+	private val applier = MosaicNodeApplier(::doLayout)
 	val rootNode = applier.root
 	private val recomposer = Recomposer(composeContext)
 	private val composition = Composition(applier, recomposer)
 
+	private val applyObserverHandle: ObserverHandle
+
+	private val readingStatesOnLayout = mutableSetOf<Any>()
+	private val readingStatesOnDrawing = mutableSetOf<Any>()
```
JakeWharton: We might want to use the new collections inside androidx.collection to get more efficiency from this in the future. Can be a follow-up (and I'll add it to #190).

JakeWharton (approved):
I think I mostly follow this 🤐

gradle/libs.versions.toml
```diff
@@ -11,6 +11,8 @@ kotlin-test = { module = "org.jetbrains.kotlin:kotlin-test", version.ref = "kotl
 kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "kotlinx-coroutines" }
 kotlinx-coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "kotlinx-coroutines" }
 
+androidx-collection = "androidx.collection:collection:1.5.0-alpha01"
```
EpicDima: I don't know why, when using version `1.4.3` (the latest stable) gradle sync closes with the error `transformNativeMainCInteropDependenciesMetadataForIde` for some modules.

mosaic-terminal/build.gradle
```diff
@@ -14,7 +14,7 @@ kotlin {
 		compilations.main.compileTaskProvider.configure {
 			doFirst {
 				def files = project.fileTree('src/jvmMain/resources').files
-				if (files.size() != 6) {
+				if (files.size() != 7) {
```
EpicDima: There was an error, as 7 files were generated, probably after adding `risc` to build native artifacts.

EpicDima:
> Is this derived from what Compose UI is doing?

It seems that the Compose UI does the same thing in a more advanced way. And also at the level of individual nodes.

mosaic-terminal/build.gradle
JakeWharton: Hmm weird. Why didn't that PR fail?

JakeWharton: I guess it was cached because the inputs didn't change. I need to hook into the process resources task or whatever instead.

JakeWharton: #476

JakeWharton:
Hmm haven't seen this failure before. We've had ID signature clashes, but they were always on the JS target not native (https://github.com/JetBrains/compose-multiplatform/issues/3418).

JakeWharton:
Okay so we're getting JetBrains' repackaging of collection as well as the AndroidX one now. I tried to use dependency substitution tricks, but with each platform being its own Maven coordinate it's gets annoying really quickly. I'm just going to back out that change for now and we can come back to it once JetBrains updates or Google starting shipping MPP runtime themselves.

PR #479, opened 2024-09-17
Send real frame times into Compose

### LINKED ISSUE #429: Send real frame time
It's needed for animations and such

JakeWharton:
Closes #429 

---

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
@@ -70,12 +70,16 @@ public fun runMosaicBlocking(content: @Composable () -> Unit) {
 }
 
 public suspend fun runMosaic(content: @Composable () -> Unit) {
+	runMosaic(enterRawMode = true, content)
+}
+
+internal suspend fun runMosaic(enterRawMode: Boolean, content: @Composable () -> Unit) {
 	val mordantTerminal = MordantTerminal()
 	val rendering = createRendering(mordantTerminal.terminalInfo.ansiLevel.toMosaicAnsiLevel())
 	val terminalState = mordantTerminal.toMutableState()
 	val keyEvents = Channel<KeyEvent>(UNLIMITED)
 
-	val rawMode = if (MultiplatformSystem.readEnvironmentVariable("MOSAIC_RAW_MODE") != "false") {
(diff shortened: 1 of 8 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -70,12 +70,16 @@ public fun runMosaicBlocking(content: @Composable () -> Unit) {
 }
 
 public suspend fun runMosaic(content: @Composable () -> Unit) {
+	runMosaic(enterRawMode = true, content)
+}
+
+internal suspend fun runMosaic(enterRawMode: Boolean, content: @Composable () -> Unit) {
```
JakeWharton, line 76: I need `awaitComplete` on the tester, but what I really need is to go to bed so punting on that until tomorrow...

mosaic-runtime/src/linuxMain/kotlin/com/jakewharton/mosaic/platform.kt
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
```
JakeWharton, line 2: All copied from Molecule

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/platform.kt
```diff
@@ -9,3 +9,5 @@ internal expect inline fun AtomicBoolean.set(value: Boolean)
 internal expect inline fun AtomicBoolean.compareAndSet(expect: Boolean, update: Boolean): Boolean
 
 internal expect inline fun atomicBooleanOf(initialValue: Boolean): AtomicBoolean
+
+internal expect inline fun nanoTime(): Long
```
EpicDima, line 13: Alternatively, it was possible to use [TimeMark](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/-time-mark/) from Kotlin, which under the hood will use the most accurate time available in the system. There would be a little more code in commonMain, but without the code in the platform source sets.

JakeWharton: By capturing an instance and then calling `elapsedNow().inWholeNanoseconds`? My problem with that is it has to make this low-level call to get the current time and then do the substraction, pack the result into a double, and then we unpack the double. I'd rather just make the call directly.

Now once kotlinx.datetime's `Instant` moves to the stdlib... maybe we can switch to that and eliminate all this platform stuff.

PR #509, opened 2024-11-01
Run on more Mac runners

JakeWharton:
---

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -10,8 +10,8 @@ on:
       - '**'
 
 jobs:
-  jni-build:
-    runs-on: macos-latest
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yaml
```diff
@@ -35,7 +35,9 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-latest
+          - macos-15 # ARM
+          - macos-14 # ARM
```
mattprecious: Why is two ARMs useful?

JakeWharton: Mostly just a hedge against Apple breaking/changing any of the OS APIs we're interacting with.

PR #514, opened 2024-11-02
Create simple Snake game sample

EpicDima:
<details>
<summary>Screenshot</summary>
Screenshot 2024-11-03 at 02 04 11
</details>

---

### DIFF
--- a/samples/snake/src/main/kotlin/example/SnakeUiAction.kt
+++ b/samples/snake/src/main/kotlin/example/SnakeUiAction.kt
@@ -1,0 +1,14 @@
+package example
+
+import example.models.Direction
+
+sealed interface SnakeUiAction {
+
+	object SwitchPhase : SnakeUiAction
+
+	class ChangeDirection(val newDirection: Direction) : SnakeUiAction
+
+	object Exit : SnakeUiAction
+
+	object Restart : SnakeUiAction
+}
--- a/samples/snake/src/main/kotlin/example/SnakeViewModel.kt
+++ b/samples/snake/src/main/kotlin/example/SnakeViewModel.kt
@@ -1,0 +1,292 @@
+package example
+
+import example.models.Direction
+import example.models.Food
+import example.models.GameState
+import example.models.INVALID_FOOD
+import example.models.Phase
+import example.models.Point
+import example.models.Size
+import example.models.Snake
+import example.models.copy
+import example.models.head
+import example.models.height
+import example.models.rawHead
+import example.models.width
+import example.models.x
+import example.models.y
+import example.utils.asLong
+import java.util.concurrent.atomic.AtomicReference
+import kotlin.random.Random
+import kotlin.system.exitProcess
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+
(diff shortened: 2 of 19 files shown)


JakeWharton:
Amazing stuff, as always. I'll fix the test stuff.

samples/snake/src/main/kotlin/example/SnakeUiAction.kt
```diff
@@ -0,0 +1,14 @@
+package example
+
+import example.models.Direction
+
+sealed interface SnakeUiAction {
+
+	data object SwitchPhase : SnakeUiAction
+
+	class ChangeDirection(val newDirection: Direction) : SnakeUiAction
```
JakeWharton, line 9:
```suggestion
	data class ChangeDirection(val newDirection: Direction) : SnakeUiAction
```
Not data?

samples/snake/src/main/kotlin/example/utils/PackingUtils.kt
```diff
@@ -0,0 +1,17 @@
...
+	return packInts(first, second)
+}
+
+inline fun packInts(val1: Int, val2: Int): Long {
+	return (val1.toLong() shl 32) or (val2.toLong() and 0xFFFFFFFF)
+}
+
+inline val Long.first: Int get() = (this shr 32).toInt()
+
+inline val Long.second: Int get() = (this and 0xFFFFFFFF).toInt()
```
JakeWharton: These could use slightly more descriptive names. Maybe call these `lowerInt` and `upperInt`, and then name the parameters of the pack function 'lower' and 'upper'.

samples/snake/src/main/kotlin/example/SnakeViewModel.kt
```diff
@@ -0,0 +1,286 @@
...
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+
+private const val INITIAL_SNAKE_SIZE = 3
+
+class SnakeViewModel(
+	private val fieldSize: Size = Size(60, 20),
+	private val updateDelayMs: Long = 200L,
+) {
+	private val coroutineScope = CoroutineScope(Dispatchers.Default)
```
JakeWharton: This should probably be injected in the constructor from a `rememberCoroutineScope` by the caller. We can use a child scope within here with its own job that can be cancelled.

samples/snake/src/main/kotlin/example/SnakeViewModel.kt
```diff
@@ -0,0 +1,286 @@
...
+		.onEach { checkGameOver(it) }
+		.map { it.toUiState() }
+		.distinctUntilChanged()
+		.stateIn(coroutineScope, SharingStarted.Lazily, gameState.get().toUiState())
+
+	private var gameCycleJob: Job? = null
+
+	init {
+		coroutineScope.launch {
+			launch {
```
JakeWharton: Remove double launch

samples/snake/src/main/kotlin/example/SnakeViewModel.kt
```diff
@@ -0,0 +1,286 @@
...
+		extraBufferCapacity = 1,
+		onBufferOverflow = BufferOverflow.DROP_LATEST,
+	)
+
+	val uiStateFlow: StateFlow<SnakeUiState> = updateUiStateFlow
+		.map { gameState.get() }
+		.onEach { checkGameOver(it) }
+		.map { it.toUiState() }
+		.distinctUntilChanged()
+		.stateIn(coroutineScope, SharingStarted.Lazily, gameState.get().toUiState())
```
JakeWharton: In general this isn't my favorite state management. It's a bit of a state soup. I would think we can put a bit more directly into composables rather than pulling it completely out. Can deal with it later.

JakeWharton (approved):

samples/snake/src/main/kotlin/example/utils/PackingUtils.kt
EpicDima: The idea of these properties is to get the first and second values from `Long` (the internal representation of `IntIntPair`), although yes, it looks like a common utility thing. I renamed it to match my idea.
In general, these extensions are needed, since in our version `Long` inside `IntIntPair` is still internal.

samples/snake/src/main/kotlin/example/SnakeUiAction.kt
EpicDima: Yes, this is not necessary in this case, I will also remove data from objects (by default, Android Studio asks to specify, now I have disabled this inspection)

PR #523, opened 2024-11-13
Use a separate wait handle from the read handle

JakeWharton:
This allows our test writer to use an anonymous pipe plus a separate event to send data. Using only the pipe does not work, as it is always in the signaled state.

---


JakeWharton:
These damn caches. Wtf.

Issue #507, opened 2024-10-31
Render as fast as possible


JakeWharton:
Right now we render every 50ms, but in reality we need to go as fast as possible. We can send frames endlessly, but they should do minimal work unless something has changed. And they absolutely shouldn't re-render _unless_ something has changed.

EpicDima:
Maybe we can change `delay(50L)` to `yield` in the loop in `mosaic.kt`?

It might also be useful to create a benchmark module to track performance with different changes

JakeWharton:
Yeah, great idea.

JakeWharton:
I recently updated [Redwood](https://github.com/cashapp/redwood) rendering and have two observations:
- The applier start/end change callbacks are a good indicator of changes that need rendered, but cannot be relied upon for _when_ to render. This is because things like `movableContentOf` will actually perform two passes over the node tree and produce two start/end callback pairs.
- Rendering should occur at the end of a frame by the thing which sends the frame. There is no first frame, however, and so at the end of `setContent` there needs to be a render.

PR #528, opened 2024-11-17
Create `mosaic-testing` module

### LINKED ISSUE #4: Testing support

EpicDima:
Fixes #4 

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 New:
 - Create `mosaic-animation` library, that provides various possibilities for animating Mosaic. An analog of [androidx.compose.animation-core](https://developer.android.com/reference/kotlin/androidx/compose/animation/core/package-summary).
 - Add `IntrinsicSize` and related `Modifier.width/height/requiredWidth/requiredHeight`.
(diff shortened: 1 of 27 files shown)


JakeWharton:
I really want to land this, but I want to play around with the API locally first. I just haven't had a lot of free time lately. I will try to get to it some day this week, hopefully tomorrow.

I think we can avoid exposing internal API for this by making the base testing API a little more generic. This would allow the normal testing API to still provide consuming of the rendered output, but the commonTest of the runtime could still consume the internal nodes.

JakeWharton:
Just a rebase so I can start looking locally. I'm off for two weeks after tomorrow, so hoping to get some movement here.

At the very least, the test API needs changed to track the changes in Redwood and #507 where we get rid of the channel and capture changes after a call to `sendFrame`.

JakeWharton:
Note: still very broken, but closer. I want to pull out a few more changes ahead of this. We're close!

PR #529, opened 2024-11-17
Add `DrawScope.drawCircle`

EpicDima:
<details>
<summary>A couple of pictures</summary>

</details>

---

- [X] `CHANGELOG.md`'s "Unreleased" section has been updated, if applicable.


JakeWharton:
So this looks fine as far as the code goes.

I guess my only question is... would someone actually use this? Circles on a monospace grid don't seem all that useful.

I could see someone using braille characters (a little like https://chipaca.com/en/2024/01/xbmbraille/), but I don't want to provide that in the core. I'd be open to another module with braille-shape drawing primitives, though.

PR #575, opened 2024-12-14
Use a pseudo console for Windows StdinWriter

JakeWharton:
---


JakeWharton:
I have convinced myself that it is impossible to fake a `Console` such that `ReadConsoleInput` works (which is the ultimate goal of this change).

Wiiindoooows \*shakes fist*.

Probably just going to manually test this (when I add it) and hope it never breaks, sticking only to VT sequence tests otherwise. I'm actually not even sure we can test the SIGWINCH/ioctl on Linux and macOS, but I _knew_ that Windows would probably be a bigger challenge/blocker.

PR #578, opened 2024-12-19
Render as fast as possible

### LINKED ISSUE #507: Render as fast as possible

JakeWharton:
Do not delay to space out frames. Instead, send frames as fast as possible and only render when something changes. This ensures we reflect changes as soon as possible rather than after an artificial delay.

Closes #507 

@EpicDima Please review if you have time!

---

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
@@ -170,7 +170,7 @@ internal class MosaicComposition(
 	private val composeContext: CoroutineContext = coroutineScope.coroutineContext + job + clock
 	val scope = CoroutineScope(composeContext)
 
-	private val applier = MosaicNodeApplier(::performLayout)
+	private val applier = MosaicNodeApplier { needLayout = true }
 	val rootNode = applier.root
 	private val recomposer = Recomposer(composeContext)
 	private val composition = Composition(applier, recomposer)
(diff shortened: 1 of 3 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -170,7 +170,7 @@ internal class MosaicComposition(
 	private val composeContext: CoroutineContext = coroutineScope.coroutineContext + job + clock
 	val scope = CoroutineScope(composeContext)
 
-	private val applier = MosaicNodeApplier(::performLayout)
```
JakeWharton, line 173: This behavior was flat out wrong. I also made the same mistake in Redwood.

If you use `movableContentOf` and move something within a single recomposition (such as from within `Row { }` to a `Column { }`), you'll actually get two pairs of calls to begin/end where the former performs the node removal and the latter performs the node add. If you render solely on the after callback, you'll render an invalid state which is missing all of the content from `movableContentOf`s.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/mosaic.kt
```diff
@@ -213,28 +212,19 @@ internal class MosaicComposition(
...
+					if (state in readingStatesOnLayout) {
+						needLayout = true
+						break
+					}
+					if (setDraw) {
+						continue
+					}
+					if (state in readingStatesOnDraw) {
+						setDraw = true
+						needDraw = true
```
JakeWharton, line 227: This should be functionally equivalent, but with early-exits where possible and minimizing volatile reads and writes.

EpicDima:
Thank you, everything is great for me!

ahmadtahagad:
Hi Jake,

I hope this message finds you well. I admire the amazing work you’ve done in the software development community, and I want to share something deeply important to me.

This may be the weirdest way anyone has ever reached out to you, but it’s coming from a sincere place. I want a beautiful life for you after death. This world, with all its achievements and wonders, is ultimately a temporary place. Everything we do, everything we create, all fades in comparison to the greatness of our Creator—Allah (God).

Islam teaches that true peace and purpose come from worshiping Allah in the way He has commanded. It’s not just about rituals, but about building a personal connection with Him, following His guidance, and preparing for what comes after this life.

I want to see you in Jannah (Paradise), a place of unimaginable peace, joy, and fulfillment. It’s incredibly precious and, yes, costly—entry requires belief in Allah and being a Muslim. I’d love to share more if you’re open to it.

Wishing you all the best in this life and the next!

Warm regards,

PR #598, opened 2025-01-02
Defer rendering in tests unless requested

JakeWharton:
This is a step toward decoupling the testing framework from the snapshot mechanism.

---


JakeWharton:
This can't work. We need to perform the actual rendering within the `onDraw` callback so that read state tracking is performed. Yay, tests!

PR #614, opened 2025-01-03
Change generic `assertFails` to `assertFailure` with specific exception

### LINKED ISSUE #611: Remove bare `assertFails` in `OffsetTest`
...and anywhere else.

At the very least we should _always_ be asserting on a specific exception type, but ideally also a message. This ensures we are testing the failure mode that we expect, and not masking larger problems.

EpicDima:
Fixes https://github.com/JakeWharton/mosaic/issues/611

---

### DIFF
--- a/mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/layout/OffsetTest.kt
+++ b/mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/layout/OffsetTest.kt
@@ -4,13 +4,13 @@ import assertk.assertThat
 import assertk.assertions.isEqualTo
 import com.jakewharton.mosaic.TestChar
 import com.jakewharton.mosaic.TestFiller
+import com.jakewharton.mosaic.assertFailure
 import com.jakewharton.mosaic.modifier.Modifier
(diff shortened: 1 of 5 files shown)


mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/layout/OffsetTest.kt
```diff
@@ -41,7 +41,8 @@ class OffsetTest {
 					TestFiller(modifier = Modifier.size(1))
 				}
 			}
-			assertFails {
+			// check when going beyond the border of the TextSurface#get
```
EpicDima, line 44: We should probably change the logic inside `TextSurface#get`, because otherwise using `Modifier.offset` is dangerous due to crashes, which is not present in the same cases in the Compose UI, for example. 
We can do as in https://github.com/JakeWharton/mosaic/pull/255 or in another way.

JakeWharton:
I will rebase this after my next PR which is a _big_ test migration.

JakeWharton:
I switched to a custom `assertFailure` rather than use kotlin.test. Still pushing for https://github.com/willowtreeapps/assertk/issues/511.

PR #617, opened 2025-01-04
Add `Modifier.composed`

EpicDima:
It may be useful to someone.

<details>
<summary>Some text</summary>

I tried to use it when trying to adapt `ExterExitTransition`, `AnimationModifier`, `AnimatedVisibility`, where the `Modifier.compose` was used, but there are difficulties with them besides the lack of`Modifier.compose`, for example, `AnimationModifier` uses `CoroutineScope` inside the `Modifier` to start animation, as well as with `Modifier.offset`, there is going beyond the boundaries during animation, the absence of lookahead, also fade cannot be adapted and scale. If I get around all this, then I get an unsatisfactory result, so I think it will wait for a bright future.
</details>
---


JakeWharton:
I've never used this in Compose UI. Where can I find some examples of its usefulness? I figure since we don't have modifier nodes (and don't plan to) maybe it's actually more useful than it has become in Compose UI, but I'd still like to know why.

EpicDima:
As one example, I don't know how much it should be implemented this way, but at least a real-life example.

If semantic colors are created in the application using the design system, then there are colors that, depending on the light or dark theme, will be different hex colors.
For example, `Background` will be white during a light theme, and black during a dark theme.

Also, the color may differ from some other conditions, for example, just a larger number of themes, different `CompositionLocal`.

This class can be used in different project `Modifier`s.

In the code, it will look something like this (simplified):
```kotlin
sealed interface DesignSystemColor {
    @JvmInline
    value class Fixed(val value: Color) : DesignSystemColor
    
    enum class Semantic : DesignSystemColor {
       Background,
       Foreground,
       // ...
    }
}

@Composable
fun DesignSystemColor.color(): Color {
  return when (this) {
    is Fixed -> this.value
    is Semantic -> {
      when (this) {
        Semantic.Background -> {
           if (isSystemInDarkTheme()) { // <- Composable
              Color.BLACK
           } else {
              Color.WHITE
           }           
        }
        Semantic.Foreground -> {
           // ...
        }
    }
  }
}

fun Modifier.background(color: DesignSystemColor): Modifier {
    return composed {
       this.background(color.color(), RoundedCornerShape(8))
    }
}
```

`Modifier.compose` can also be used, I'm also not sure if this is the most correct way, to create `Modifier`s that simply simplify life without forcing you to pass a bunch of parameters to them.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ New:
 - Create `mosaic-animation` library, that provides various possibilities for animating Mosaic. An analog of [androidx.compose.animation-core](https://developer.android.com/reference/kotlin/androidx/compose/animation/core/package-summary).
 - Add `IntrinsicSize` and related `Modifier.width/height/requiredWidth/requiredHeight`.
 - New `mosaic-testing` artifact for testing Mosaic.
+- Add `Modifier.composed`.
```
JakeWharton: Please add at least a full sentence here about what it's for. Can be copied from the docs or something.

JakeWharton (approved):

EpicDima:
Now there is a `Modifier` in the Compose UI and Compose Multiplatform there is `Modifier.Node`, through which you can do many things that were previously implemented through `Modifier.composed`, so `Modifier.composed` is now less frequently used, as it seems to me.

JakeWharton:
Yeah I'm hoping we won't need nodes. Both the number of tree nodes and the number of modifiers applied to each is dramatically lower than Compose UI. Plus we're running on the JVM/native which should be faster. Plus our tree nodes and the layers produced from modifier elements are so much simpler and faster than Compose UI.

🤞

EpicDima:
I've never thought about it that way (It's about simplicity and thus speed, which is really the case). In the release notes Compose UI, the guys wrote that they had refactored `Modifier`s by creating a `Modifier.Node`, so that processing is faster. I wonder why they have such problems? Maybe the `Modifier.Node` approach will be better in the future?

JakeWharton:
When a function recomposes the modifiers are all recreated. By switching to nodes, the modifiers become factories of objects which do work rather than objects which do work themselves. This means the allocations are smaller, and then the list of them can be diffed against the old list with only the necessary changes happening.

PR #619, opened 2025-01-04
Create internal `Modifier.testTag`

EpicDima:
I've made it internal at the moment, since `NodeSnapshots` is still inside `:mosaic-runtime`, and therefore `Modifier.testTag` simply cannot be used outside of this module without some kind of extension of the `Mosaic` interface, which I'm not sure about right now. 
This can be done through the extension function, which will check the `Mosaic` type for `MosaicComposition` and then search for the desired node...

The Compose UI uses `SemanticsModifier` and a number of other things that are used not only for `TestTag`, but also many other things. The most interesting thing for us there besides `TestTag` is `Text`, so in order not to drag a bunch of code, which is not at all necessary right now, plus integration is not easy, I decided to do it in a simple way.

---


JakeWharton:
Haven't looked at the PR yet, but we'll eventually need some kind of Espresso like testing API so we'll need this for node selection. I have no plans to expose any kind of node directly, though. I don't believe we need a public semantic tree either.

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/stuff.kt
```diff
@@ -48,6 +48,18 @@ internal val MosaicNode.position: IntOffset
 
 class Holder<T>(var value: T)
 
+internal fun MosaicNode.nodeByTestTag(tag: String): MosaicNode {
+	val queue = mutableListOf(this@nodeByTestTag)
+	while (queue.isNotEmpty()) {
+		val node = queue.removeAt(0)
```
JakeWharton: "deque" == double-ended queue, i.e., efficient appends and removes from both ends.
```suggestion
	val queue = ArrayDeque().also { it.add(this) }
	while (queue.isNotEmpty()) {
		val node = queue.removeFirst()
```

JakeWharton (approved):

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/stuff.kt
JakeWharton: This might fail to build because of lack of generics, but hoping K2 can figure it out.

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/stuff.kt
```diff
@@ -48,6 +48,18 @@ internal val MosaicNode.position: IntOffset
 
 class Holder<T>(var value: T)
 
+internal fun MosaicNode.nodeByTestTag(tag: String): MosaicNode {
+	val queue = ArrayDeque().also { it.add(this) }
```
JakeWharton:
```suggestion
	val queue = ArrayDeque<MosaicNode>().also { it.add(this) }
```

mosaic-runtime/src/commonTest/kotlin/com/jakewharton/mosaic/stuff.kt
JakeWharton: Yep. Failed.

https://youtrack.jetbrains.com/issue/KT-40346

Issue #624, opened 2025-01-05
Beef up root and module's `README.md` and add to Dokka


JakeWharton:

JakeWharton:
I think we can re-land this now that the 'application' plugin has been fully jettisoned.

PR #625, opened 2025-01-05
Add README's to dokka

EpicDima:
- Move the gifs from separate directories to images so that the pictures were pulled up in the documentation, there is also a folder called images
- Rename the gifs by the name of the sample from which it was created, since the same names cannot be in the same folder
- Update the script for creating gifs for updating in a new directory
- Update README.md for samples for pulling up gifs from the images root directory
- In the root README.md add a width for images, because without it, the images in the documentation stretch to the full width of the page
- Add a simple README.md to the `mosaic-testing` and `mosaic-runtime` modules
- Update the header in README.md the `mosaic-animation` and `mosaic-terminal` modules for correct processing (https://kotlinlang.org/docs/dokka-module-and-package-docs.html#file-format)
- Add saving to assets for dokka of all images from the root README.md automatically
- Change `dokkaSourceSets.named('commonMain').configure` -> `dokkaSourceSets.configureEach` to correctly display information from README.md for different targets (common, jvm, native)

Solution for https://github.com/JakeWharton/mosaic/issues/624
Fix for https://github.com/JakeWharton/mosaic/issues/627 (not really honest, more like a workaround)

---


JakeWharton (approved):
Nice! Looks good (from mobile). Can you put `images/` beneath `docs/`, just to keep the top-level more clean.

I'll check the rendered docs when I get home.

EpicDima:
> Can you put images/ beneath docs/, just to keep the top-level more clean.

Unfortunately, the problem is that `customAssets` cannot be nested, meaning that after rendering, all images are placed inside the `images` directory without subfolders. And so that the images are displayed when opening `README.md` and after the documentation rendering, I had to do it exactly at the top level.

JakeWharton:
I'd like to explore what's going on in Dokka a bit more before landing this. Can we simply revert to the v1 rendering for now? Also seems like I need to run more of the samples on CI...

PR #626, opened 2025-01-05
Add CODE_OF_CONDUCT file

Marwane-20:
Hello,
Just noticed there was no CODE_OF_CONDUCT file, and i added one from the https://www.contributor-covenant.org/. 

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,133 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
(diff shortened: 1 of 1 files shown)


JakeWharton:
I'm sure this is fine, but I need to actually read it first and I'm too tired right now. Will do tomorrow.

EpicDima:
Alternatively, we can put the file in `.github`, like here for example https://github.com/square/logcat/blob/main/.github/CODE_OF_CONDUCT.md (although I don't know the reasons why it is there)

JakeWharton:
Yeah these are called community health files. https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/about-community-profiles-for-public-repositories

They can be in the root, in docs/, or in .github/.

Marwane-20:
Indeed, the majority of repos that I see, have it in root, and sometimes in docs/, but that's just my experience.

JakeWharton:
Looks fine. Thanks!

PR #647, opened 2025-01-16
Use development versions of KGP

JakeWharton:
I want to test new build features as they're developed.

Need to first:
* [ ] Diff poms to ensure nothing has changed
* [ ] Diff metadata to ensure nothing has changed

---

### DIFF
--- a/build-support/build.gradle
+++ b/build-support/build.gradle
@@ -7,6 +7,9 @@ buildscript {
 	}
 	repositories {
 		mavenCentral()
+		maven {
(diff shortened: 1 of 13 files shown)


JakeWharton:
Maybe it's better to resolution strategy the stdlib instead of trying to add it ourselves?

JakeWharton:
This approach isn't going to work. JS and Wasm want the same version of the stdlib as the compiler, and native sets the klib version based on the compiler (ignoring the language and api version set).

PR #649, opened 2025-01-16
Grow some ARMs on CI

JakeWharton:
---


JakeWharton:
Oh, right. Kotlin toolchain can't run on Linux ARM (as I am _intimately_ aware). We can copy the jars and kexe over, there's really no need to build on every target (we only do it because Gradle makes running and failing easier, but we don't have to).

PR #650, opened 2025-01-16
Build tests once and copy to other platforms to run

JakeWharton:
---


build-support/src/main/kotlin/com/jakewharton/mosaic/buildsupport/MosaicBuildExtensionImpl.kt
```diff
@@ -0,0 +1,154 @@
...
+
+	interface GradleSupport {
+		fun detectTestClassNames(
+			testClasses: FileTree,
+			testClassDirectories: List<File>,
+			testClasspath: List<File>,
+		): List<String>
+	}
+
+	class Gradle_8_10_Support : GradleSupport {
```
JakeWharton, line 125: We're on Gradle 8.12 now, but when I originally wrote this it worked on Gradle 8.10...

PR #651, opened 2025-01-22
Rewrite native input handling

JakeWharton:
This will allow out-of-band events to come in, such as a resize from SIGWINCH, or focus and resize from Windows' console events.

---


JakeWharton:
Tempted to just merge this despite the fact that the naming is a mess and there's duplication and other annoyances. It does work, however....

JakeWharton:
Lots to still do here, but I'd rather land and do small changes to get where I want. The architecture change is the most important thing.

PR #806, opened 2025-03-18
Bring back Kitty underlines!

### LINKED ISSUE #802: `Text` composable broken on native
Started with #765 (bisected to 8957294eff99f6ca3b744b12e146b8131ac1b3ad). Not sure why, but probably Compose compiler bug.

JakeWharton:
The difference in `UnderlineStyle` as opposed to `Color` and `TextStyle` was that it had a `private` constructor. Changing that to `internal` like the others avoids the compiler bug.

Closes #802 

---


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/ui/UnderlineStyle.kt
```diff
@@ -8,7 +8,7 @@ import kotlin.jvm.JvmInline
 
 @Immutable
 @JvmInline
-public value class UnderlineStyle private constructor(
+public value class UnderlineStyle internal constructor(
```
mattprecious, line 12: Worth commenting that this cannot be private due to the bug?

JakeWharton: Sure. I did also change the build configuration so it will fail if you try.

PR #826, opened 2025-03-25
Build all samples on their own

JakeWharton:
This does not yet exclude them from the main build.

---


JakeWharton:
This seems like too much wasted effort unless we first build all of the main modules and then fan out to the samples. Not going to mess with it.

PR #858, opened 2025-04-13
More TTY terminal capability testing

JakeWharton:
---


JakeWharton:
Still need to test their state flows, but already unhappy with these tests. Need Burst to test all the different mode states that could be replied.

PR #871, opened 2025-04-29
Always enable in-band resize mode even if already set

JakeWharton:
This should cause the terminal to emit the current size.

---


mosaic-tty-terminal/src/commonMain/kotlin/com/jakewharton/mosaic/tty/terminal/TtyTerminal.kt
```diff
@@ -297,7 +297,11 @@ public suspend fun Tty.asTerminalIn(
 				}
 
 				is ResizeEvent -> {
-					size.value = Terminal.Size(event.columns, event.rows, event.width, event.height)
+					if (inBandResizeEvents) {
```
JakeWharton, line 300: Actually this will end up ignoring events if we fall back to the platform.

JakeWharton: Ah, no it won't. We write directly to the state and then emit the event.

PR #877, opened 2025-05-07
Fix path issues in Zig and Gradle scripts

fluxxion82:
I have two issues related to the zig build script.

First, I had a issue with zig prefix build path being presented in the build script error message. Original path was the jni path and I needed to run the zig build command from the jni parent. Updated gradle script to ouput jni parent path in the error message.

Second issue was a command in zig build script. I have zig version 0.14.0 and I think it's more strict about build paths. I updated zig build script to use `.cwd_relative` option for the install path removal step. Not sure if this works on older versions of zig


JakeWharton:
Thanks!

PR #1017, opened 2025-10-07
Write to the stdout interrupter, ensure native is initialized in FFM `StandardStreams`

JakeWharton:
---


JakeWharton:
I don't understand why tests didn't catch the interrupt thing

PR #1068, opened 2026-01-14
Use Kotlin builtin AtomicBoolean

Goooler:
Since Kotlin 2.1

---


JakeWharton:
https://youtrack.jetbrains.com/issue/KT-82228 seems to indicate there's a very slow path to making this stable. We may have to revert this at some point (such as for a 1.0), unless we take just as long as they do...

Goooler:
Good to know!

PR #1085, opened 2026-02-11
fix crashing on common Latin language characters in input outside the unicode range 32..126

eidam-slices:
This PR fixes an issue in KeyboardEvent.toKeyEventOrNull() where pressing a key with a Unicode codepoint outside the range 32..126 would throw an UnsupportedOperationException, causing the TUI to crash.
I kept the exception throwing and added support for all Latin characters, with expansion friendly object for known Unicode blocks.
---

Changes:
	- Updated toKeyEventOrNull() to support all european Latin characters instead of throwing exception
	- Introduced a more robust Unicode codepoint conversion with toCodepoint() and UnicodeBlock utilities.
	- Added support for extended Latin ranges to cover common key inputs.

Impact:
	- Improves stability for interactive Mosaic TUI applications.
	- Preserves existing behavior for standard keys (Enter, Backspace, Arrow keys, function keys).
	- Future-proofing for additional Unicode blocks (Greek, Cyrillic, etc.) can be added easily.

Testing:
	- Verified that keypresses outside standard Latin ranges no longer throw exceptions.
	- Confirmed that normal key inputs and control keys (Enter, Backspace, arrows) continue to function correctly.

Next Steps / Recommendations:
	- Consider extending UnicodeBlocks to include Greek, Cyrillic, and other commonly used scripts for broader TUI support.
	- Optional: log or ignore unknown codepoints instead of throwing exceptions to further improve robustness.

### DIFF
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/Codepoint.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/Codepoint.kt
@@ -1,0 +1,89 @@
+package com.jakewharton.mosaic
+
+import kotlin.jvm.JvmInline
+
+internal object UnicodeBlocks {
+
+	// region Latin
+
+	val LatinBasic = Codepoint("U+0000")..Codepoint("U+007F")
+	val Latin1Supplement = Codepoint("U+00A0")..Codepoint("U+00FF")
+	val LatinExtendedA = Codepoint("U+0100")..Codepoint("U+017F")
+	val LatinExtendedB = Codepoint("U+0180")..Codepoint("U+024F")
+	val LatinExtendedC = Codepoint("U+2C60")..Codepoint("U+2C7F")
+	val LatinExtendedD = Codepoint("U+A720")..Codepoint("U+A7FF")
+	val LatinExtendedE = Codepoint("U+AB30")..Codepoint("U+AB6F")
+	val LatinExtendedF = Codepoint("U+10780")..Codepoint("U+107BF")
+	val LatinExtendedG = Codepoint("U+1DF00")..Codepoint("U+1DFFF")
+	val LatinExtendedAdditional = Codepoint("U+1E00")..Codepoint("U+1EFF")
+
+	val LatinIpaExtensions = Codepoint("U+0250")..Codepoint("U+02AF")
+	val PhoneticExtensions = Codepoint("U+1D00")..Codepoint("U+1D7F")
+	val PhoneticExtensionsSupplement = Codepoint("U+1D80")..Codepoint("U+1DBF")
+
+	val SpacingModifierLetters = Codepoint("U+02B0")..Codepoint("U+02FF")
+	val CombiningMarks = Codepoint("U+0300")..Codepoint("U+036F")
+
+	val latin = listOf(
+		LatinBasic,
+		Latin1Supplement,
+		LatinExtendedA, LatinExtendedB, LatinExtendedC,
+		LatinExtendedD, LatinExtendedE, LatinExtendedF,
+		LatinExtendedG,
+		LatinExtendedAdditional,
+		LatinIpaExtensions, PhoneticExtensions, PhoneticExtensionsSupplement,
+		SpacingModifierLetters, CombiningMarks
+	)
+	// endregion
+
+	// TODO: Greek, Cyrillic, ...
+}
+
+@JvmInline
+internal value class Codepoint constructor(val dec: Int) {
+
+	constructor(unicode: String) : this(unicode.removePrefix(PREFIX).toInt(16))
+
+	val hex: String get() = dec.toString(16).uppercase()
+	val unicode: String get() = PREFIX + hex
+
+	companion object {
+		const val PREFIX = "U+"
+	}
+
+	fun toChar(): Char = dec.toChar()
+
+	override fun toString(): String = buildString {
+		if (dec <= 0xFFFF) {
+			// simple BMP code point:
+			append(toChar())
+		} else {
+			// surrogate pair calculation:
+			// https://en.wikipedia.org/wiki/Plane_(Unicode)#Surrogate_pairs
+			val high = ((dec - 0x10000) shr 10) + 0xD800
+			val low = ((dec - 0x10000) and 0x3FF) + 0xDC00
+			append(high.toChar())
+			append(low.toChar())
+		}
+	}
+
+	operator fun rangeTo(other: Codepoint) = UnicodeBlock(start = this, endInclusive = other)
+}
+
+internal data class UnicodeBlock(val start: Codepoint, val endInclusive: Codepoint) {
+
+	operator fun contains(codepoint: Codepoint): Boolean {
+		return codepoint.dec in this.toIntRange()
+	}
+
+	fun toIntRange() = IntRange(start.dec, endInclusive.dec)
+}
+
+internal fun IntRange.toUnicodeBlock(): UnicodeBlock {
+	return UnicodeBlock(
+		start = Codepoint(start), endInclusive = Codepoint(endInclusive)
+	)
+}
+
+internal fun Int.toCodepoint() = Codepoint(this)
+internal fun String.toCodepoint() = Codepoint(this)
--- a/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt
+++ b/mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt
@@ -13,7 +13,6 @@ internal fun KeyboardEvent.toKeyEventOrNull(): KeyEvent? {
 			9 -> "Tab"
 			13 -> "Enter"
 			27 -> "Escape"
-			in 32..126 -> codepoint.toChar().toString()
 			127 -> "Backspace"
 			57350 -> "ArrowLeft"
 			57351 -> "ArrowRight"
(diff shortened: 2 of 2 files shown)


mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt
```diff
@@ -26,7 +25,15 @@ internal fun KeyboardEvent.toKeyEventOrNull(): KeyEvent? {
 			57356 -> "Home"
 			57357 -> "End"
 			in 57364..57398 -> "F" + (codepoint - 57363)
-			else -> throw UnsupportedOperationException(toString())
+			else -> {
+				val codepoint = codepoint.toCodepoint()
+
+				if (UnicodeBlocks.latin.any { codepoint in it }) {
+					codepoint.toChar().toString()
```
gabrieljones, line 32: mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt:R32

**`toChar()` truncates supplementary plane codepoints (above U+FFFF) to 16 bits**

In `compat.kt:32`, `codepoint.toChar().toString()` is called for any codepoint that falls within the Latin unicode blocks. However, `Codepoint.toChar()` at `Codepoint.kt:54` simply does `dec.toChar()`, which truncates to the lower 16 bits for codepoints above U+FFFF.

<details>
<summary>Affected Unicode blocks and root cause</summary>

The Latin blocks include Latin Extended F (U+10780–U+107BF) and Latin Extended G (U+1DF00–U+1DFFF), both of which are in the supplementary planes (above U+FFFF). For these codepoints, `Int.toChar()` only keeps the lower 16 bits, producing a completely wrong character.

For example, codepoint U+10780 (dec 67456) would become `67456.toChar()` which yields `Char(0x0780)` — a Thaana character instead of the intended Latin Extended F character.

The `Codepoint.toString()` method correctly handles surrogate pairs for supplementary plane characters, but `toChar()` does not. The fix is to use `codepoint.toString()` instead of `codepoint.toChar().toString()` on line 32.

**Impact:** Any keyboard input producing a Latin Extended F or G codepoint will be mapped to the wrong key name.

</details>

https://app.devin.ai/review/JakeWharton/mosaic/pull/1085

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/Codepoint.kt
```diff
@@ -0,0 +1,89 @@
+package com.jakewharton.mosaic
+
+import kotlin.jvm.JvmInline
+
+internal object UnicodeBlocks {
+
+	// region Latin
+
+	val LatinBasic = Codepoint("U+0000")..Codepoint("U+007F")
```
gabrieljones, line 9: mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/Codepoint.kt:R9

**LatinBasic block (U+0000–U+007F) includes C0 control characters that are now treated as printable keys**

The `LatinBasic` block is defined as `U+0000..U+007F` at `Codepoint.kt:9`, which includes C0 control characters (0–31). The `when` expression only handles a few specific control characters (Tab=9, Enter=13, Escape=27). All other control characters (0–8, 10–12, 14–26, 28–31) now fall through to the `else` branch and match `LatinBasic`, causing them to be converted to key names via `toChar().toString()`.

<details>
<summary>Regression from previous behavior</summary>

The old code used `in 32..126` which correctly excluded all control characters. The new code's `UnicodeBlocks.latin` list includes `LatinBasic` (U+0000–U+007F), so codepoints like 0 (NUL), 1 (SOH), 10 (LF), etc. will now be treated as valid printable key events instead of throwing `UnsupportedOperationException`.

For example, if a keyboard event arrives with codepoint 0 (NUL), the old code would throw. The new code would produce `KeyEvent(key = "\u0000", ...)` — a key event with a control character as the key name, which is almost certainly not the intended behavior.

**Impact:** Unexpected control character codepoints will silently produce malformed `KeyEvent` objects instead of being flagged as unsupported.

</details>

https://app.devin.ai/review/JakeWharton/mosaic/pull/1085

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/Codepoint.kt
```diff
@@ -0,0 +1,89 @@
...
+	constructor(unicode: String) : this(unicode.removePrefix(PREFIX).toInt(16))
+
+	val hex: String get() = dec.toString(16).uppercase()
+	val unicode: String get() = PREFIX + hex
+
+	companion object {
+		const val PREFIX = "U+"
+	}
+
+	fun toChar(): Char = dec.toChar()
```
gabrieljones, line 54: mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/Codepoint.kt:R54

**Codepoint.toString() correctly handles surrogate pairs but toChar() does not — design inconsistency**

The `Codepoint` class has two ways to convert to string representation: `toChar().toString()` (line 54) and `toString()` (lines 56-68). The `toString()` method correctly handles supplementary plane characters by computing surrogate pairs, while `toChar()` silently truncates via `Int.toChar()`. This is a latent footgun in the API design — any caller using `toChar()` on a supplementary plane codepoint gets silently wrong results. Consider either making `toChar()` throw for codepoints > 0xFFFF, or documenting its BMP-only limitation.

https://app.devin.ai/review/JakeWharton/mosaic/pull/1085

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/Codepoint.kt
```diff
@@ -0,0 +1,89 @@
+package com.jakewharton.mosaic
+
+import kotlin.jvm.JvmInline
+
+internal object UnicodeBlocks {
+
+	// region Latin
+
+	val LatinBasic = Codepoint("U+0000")..Codepoint("U+007F")
+	val Latin1Supplement = Codepoint("U+00A0")..Codepoint("U+00FF")
```
gabrieljones, line 10: mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/Codepoint.kt:R9-10

**Gap between LatinBasic and Latin1Supplement (U+0080–U+009F) is intentional per Unicode**

The `LatinBasic` block ends at U+007F and `Latin1Supplement` starts at U+00A0, leaving a gap at U+0080–U+009F. This range corresponds to C1 control characters in Unicode (e.g., CSI, OSC, etc.), which are correctly excluded from the Latin printable blocks. This is consistent with the Unicode standard block definitions and is not a bug. However, it's worth noting that any keyboard event with a codepoint in this range will throw `UnsupportedOperationException`, same as the old behavior.

https://app.devin.ai/review/JakeWharton/mosaic/pull/1085

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt
```diff
@@ -26,7 +25,15 @@ internal fun KeyboardEvent.toKeyEventOrNull(): KeyEvent? {
 			57356 -> "Home"
 			57357 -> "End"
 			in 57364..57398 -> "F" + (codepoint - 57363)
-			else -> throw UnsupportedOperationException(toString())
+			else -> {
+				val codepoint = codepoint.toCodepoint()
```
gabrieljones, line 29: mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt:R29

**Variable shadowing of `codepoint` in the else branch**

In `compat.kt:29`, `val codepoint = codepoint.toCodepoint()` shadows the outer `codepoint` (an `Int` from the `when` subject at line 12) with a `Codepoint` value class. While this is valid Kotlin and the compiler handles it correctly, it can be confusing to readers since both variables have the same name but different types (`Int` vs `Codepoint`). Consider using a different name like `cp` or `codepointObj` for clarity.

https://app.devin.ai/review/JakeWharton/mosaic/pull/1085

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt
```diff
@@ -26,7 +25,15 @@ internal fun KeyboardEvent.toKeyEventOrNull(): KeyEvent? {
...
+
+				if (UnicodeBlocks.latin.any { codepoint in it }) {
```
gabrieljones, line 31: mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt:R31

**Linear scan through Latin blocks for every key event**

The `else` branch at `compat.kt:31` calls `UnicodeBlocks.latin.any { codepoint in it }`, which iterates through all 15 Latin unicode blocks for each unrecognized key event. While this is unlikely to be a performance concern in practice (keyboard events are infrequent and the list is small), it's worth noting that the blocks could be merged into a single sorted structure for O(log n) lookup if more blocks are added in the future (as the TODO at `Codepoint.kt:39` suggests).

https://app.devin.ai/review/JakeWharton/mosaic/pull/1085

eidam-slices:
@gabrieljones Thanks for notes and reviews. With the code blocks, i pretty much just used the wiki definitions, so that's why there is a lot of these bugs.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt
eidam-slices: Okay, I use this sometimes in my codes, but I agree it may be a bad convention / habit for reading.

mosaic-runtime/src/commonMain/kotlin/com/jakewharton/mosaic/compat.kt
eidam-slices: I never did this in Kotlin. Is there some function for that, or it would have to be some custom implementation?

gabrieljones: @eidam-slices I am not a Kotlin expert. I recommend visiting https://app.devin.ai/review/JakeWharton/mosaic/pull/1085 and asking Devin that question.

